PR #28, opened 2022-06-03
Add `npm-version` support

### LINKED ISSUE #27: Git tags and semantic versioning
If possible I would avoid the use of git branches to keep track of previous versions of this library, indeed, when possible I would delete branches that are too old.

Alternatively you can use automatic tagging using the [version](https://docs.npmjs.com/cli/v8/commands/npm-version) property of `package.json`.

For example:

```
  "version": "npm run build && git add -A dist",
```

_Have a nice day,
Raruto_

Raruto:
Usage examples:

```js
npm version [<newversion> | major | minor | patch | premajor | preminor | prepatch | prerelease | from-git]
```

Reference: https://docs.npmjs.com/cli/v8/commands/npm-version

closes: https://github.com/g3w-suite/g3w-client/issues/27

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "plugins": "gulp g3w-admin-plugins-select",
     "default": "gulp default --max-old-space-size=2048",
     "test": "gulp test",
-    "cy:open": "cypress open"
+    "cy:open": "cypress open",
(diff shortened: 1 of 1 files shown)


volterra79:
@Raruto done.

PR #29, opened 2022-06-06
Dockerized development

### LINKED ISSUE #26: Make use of `engines` property inside `package.json`
It should be worth a thousand words about the recommended **npm** and **node** versions to develop, see: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

_Have a nice day,
Raruto_

Raruto:
For greater clarity I took what is needed to develop this library via docker directly from here: https://github.com/g3w-suite/g3w-suite-docker/commit/8810468ef59740ae1a17f995bd2fd6d7680c16fc \* and moved it within this repository inside the [`docker`](https://github.com/Raruto/g3w-client/tree/docker/docker) folder.

There are also other non-strictly related changes (some code refactoring), but before I go too far I send it as a draft so you can check if it's good enough already (or notify me of incorrect changes).

Closes: https://github.com/g3w-suite/g3w-client/issues/26

\* _as @wlorenzetti knows actually that script works but does not create the default user and some other fixtures_

_Raruto_


wlorenzetti:
Hi @Raruto it could be a clear solution, but I'm afraid it might become difficult to keep too many dockerization schema. I think it is best to discuss this in a call :). 

For the discussion of creating users and loading fixtures, I push a fix for it: https://github.com/g3w-suite/g3w-admin/commit/fb45e5de3f9e9824bce3d1aeb15bd61b6e7ba2e4. Please try the procedure again clearing before every docker volume and directories and let me know if you have any problems.

Raruto:
> Please try the procedure again clearing before every docker volume and directories and let me know if you have any problems.

I did some tests from scratch and now the database part seems to work, now I have some problem with the contents of the `project_data` folder (which is apparently empty for windows explorer) but for the moment nothing to worry about.

> I'm afraid it might become difficult to keep too many dockerization schema.

I was hoping that those files wouldn't change so much often, but as you see, nothing definitive 😉

_Raruto_

wlorenzetti:
@Raruto `project_data` is empty by default, so I think now is ok? For dockerization into client code, ok for me ask to @volterra79 .

Raruto:
> `project_data` is empty by default

Yes, I meant that on the wsl2 side I don't see its content even if you upload something through the g3w-suite uploader.
Thinking about it I would say it is correct, since this is not a shared folder it should not be visible directly from Windows explorer, in fact through `docker exec` I can list all that files correctly.

To understand, also on the wsl2 side I can see the folder `\tmp\shared-volume-g3wsuite-dev\project_data` (which is always empty) but I think this behavior is more related to the functioning of docker on Windows.

> For dockerization into client code, ok for me ask to @volterra79 

Ok, thank you.

_Raruto_

Raruto:
Superseded by: https://github.com/g3w-suite/g3w-client/pull/828

Which introduced local-to-remote development (Hybrid Proxy @ localhost:3000)

Ref: [g3w-client/build.js](https://github.com/g3w-suite/g3w-client/blob/09f7f84640ed34704b086ccfb9cbe64ae4465df4/build.js)

PR #30, opened 2022-06-07
Update minimum NodeJS version

### LINKED ISSUE #26: Make use of `engines` property inside `package.json`

Raruto:
**package.json**
```json
"engines" : { 
  "node" : ">=16.13.2",
  "npm" : ">=8.1.2"
}
```

---

Related to: https://github.com/g3w-suite/g3w-client/pull/29
Closes: https://github.com/g3w-suite/g3w-client/issues/26


volterra79:
@Raruto  is better :

```json
"engines" : { 
  "node" : ">=14.18.1",
  "npm" : ">=8.3.0"
}
```

Raruto:
@volterra79 ok, in the meantime I have to figure out how to update npm.

If you are already on that versions, you can also update the package-lock.json file? (so I directly download the latest versions of both and we don't mess with the published lock file).

_Raruto_

Raruto:
@volterra79

If I have not misunderstood each version of node provides an npm version by default (at least this way it works on Windows).

```sh
nvm use 14.18.1
```

```sh
node -v
  v14.18.1
```

```sh
npm -v
  6.14.15
```

Somehow it would also be possible to update npm on its own, but it is much more problematic and cumbersome than using `nvm`. Is there a particular reason why you need such a high version of npm?

volterra79:
@Raruto no reason to update to the high version of npm. In my linux installation npm v8.3.0 come with node v14.18.1. So i sugget to specify only the minimun version of node in "engines" configuration settings.

Raruto:
Are you sure you don't have any PATH rules that take precedence?

On my enviroment I have the following association:

> Node.js 14.18.1 --> npm 6.14.15

See: https://nodejs.org/en/download/releases/

volterra79:
@Raruto done

Raruto:
Tested and working also with:

**node:** v16.15.0
**npm:** v8.5.5

**g3w-client:** v3.4.0

volterra79:
@Raruto thanks

PR #32, opened 2022-06-07
Refactor guplfile.js

Raruto:
First rewrite of the gulpfile to try to make it easier for anyone to understand:

- move declare statements to top
- avoid some negative conditionals
- comment out dead code
- use multiline comments for tasks
- inline some unnecessary functions definitions

Related to: https://github.com/g3w-suite/g3w-client/pull/29

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -46,9 +46,9 @@ const client = argv.client || '';
 const client_version = (client !== '') ? 'client-'+client : 'client';
 // it used to change build minified js and css to avoid server cache
 // every time we deploy a new client version
-const versionHash = Date.now();
+// const versionHash = Date.now();
 let production = false;
-let g3w_admin = false;
+// let g3w_admin = false;
 let build_all = true;
 let g3w_admin_version = 'dev';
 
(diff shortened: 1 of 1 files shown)


volterra79:
@Raruto done. I checked and i need to push a commit to fix an issue on gulpfile regarding 'browser-sync' task on line 322:
const port = conf.localServerPort ?? 3000; to const port = conf.localServerPort || 3000;

Raruto:
@volterra79 

ok, but keep in mind that the following statement:

```js
const port = conf.localServerPort ?? 3000;
```

is equivalent to the following (which was the starting code):

```js
const port = conf.localServerPort ? conf.localServerPort : 3000;
```

and not to:

```js
const port = conf.localServerPort || 3000;
```

_or at least I think so_

volterra79:
@Raruto you are right. If run **npm run default** using 
```js
const port = conf.localServerPort ?? 3000; 
```
i receive an error

`const port = conf.localServerPort ?? 3000;
                                     ^

SyntaxError: Unexpected token '?'
    at wrapSafe (internal/modules/cjs/loader.js:1047:16)
    at Module._compile (internal/modules/cjs/loader.js:1097:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1153:10)
    at Module.load (internal/modules/cjs/loader.js:977:32)
    at Function.Module._load (internal/modules/cjs/loader.js:877:14)
    at Module.require (internal/modules/cjs/loader.js:1019:19)
    at require (internal/modules/cjs/helpers.js:77:18)
`

Raruto:
yes, there is probably a project dependency with does not yet support the nullish coalescing operator.

volterra79:
@Raruto i investigate on it

PR #34, opened 2022-06-09
first refactoring using class instead prototype and use es6 module in…

volterra79:
- Start es6 module refactoring and moved to Function prototype to class. 
- Need to use npm packages instead libraries files inside vendor folder. 
- upgrade gulp and other tool for development
- and more ....


package.json
```diff
@@ -111,9 +119,5 @@
     "vueify": "^9.4.1",
     "watchify": "^3.7.0",
     "yargs": "^8.0.2"
-  },
-  "engines" : { 
-    "node" : ">=16.13.2",
-    "npm": ">=8.1.2"
```
Raruto, line 117: @volterra79 be careful, I think there is an oversight here

src/config/i18n/index.js
```diff
@@ -3355,7 +3355,7 @@ const translations = {
       search: "Căutare",
       no_results: "Niciun rezultat găsit",
       print: "Tipărire",
-      create_print: "Tipărește",
```
Raruto: @volterra79 check out: https://github.com/g3w-suite/g3w-client/commit/73f16b90002e22ed0e98a50c1072051095987ccc

package.json
volterra79: @Raruto aligned with dev. Thanks

src/config/i18n/index.js
volterra79: @Raruto aligned with dev. Thanks

PR #38, opened 2022-06-15
add 1:200 and 1:100 print scale

tudorbarascu:
I think these should be included in the default print scales as in the cities (public utilities, cadastre etc) they're normally used for detail maps.


volterra79:
@tudorbarascu thanks

PR #41, opened 2022-06-15
ro translate print

tudorbarascu:


volterra79:
@tudorbarascu done. Thanks

PR #51, opened 2022-07-08
Consolidate `.browserlistrc` and `.babelrc` config within package.json

Raruto:


Raruto:
Note for the readers: https://github.com/g3w-suite/g3w-client/commit/10b15489fdf5ab6916b3bf4a6deb73aedbc83ed1

In order to make fully work without .babelrc, the following option should be removed from within `gulpfile.js`:

https://github.com/g3w-suite/g3w-client/blob/10b15489fdf5ab6916b3bf4a6deb73aedbc83ed1/gulpfile.js#L168

PR #113, opened 2022-08-02
Remove `src/app/core/geometry` folder 

volterra79:
**List of changes:**
- delete `src/app/core/geometry/` folder
  - delete `src/app/core/geometry/geom.js` file (use `core/utils/geo` instead)
  - delete `src/app/core/geometry/geometry.js` file  (use `core/utils/geo` instead)

- move `src/app/core/utils/geo::GeometryTypes` into `src/app/constant::GeometryTypes`
- remove temporary `geo` variable within `src/app/core/utils/geo.js`
- replace `geoutils.isMultiGeometry()` function with `Geometry.isMultiGeometry()` into `src/app/core/utils/geo.js`


Raruto:
@volterra79 this pull can break a lot of things, we should not to include it in **v3.5**

volterra79:
@Raruto i tested and all it works.

PR #125, opened 2022-08-08
fix wrong default vendor keys names within `config.template.js`

Raruto:
Fixes wrong pull: https://github.com/g3w-suite/g3w-client/pull/121

- rename `BING_API_KEY` into `bing`
- rename `GOOGLE_API_KEY` into `google`

Related to: https://github.com/g3w-suite/g3w-client/pull/111


volterra79:
Align with same keys name of initConfig.group.vendorkeys provide by g3w-admin

PR #127, opened 2022-08-08
Add Google Geocoding API 

volterra79:
Add Google api, if we have a google key enabled, to geocoding search, other than Nominatim api.

Screenshot from 2022-08-08 16-30-48

Related to: https://github.com/g3w-suite/g3w-client/pull/111

### DIFF
--- a/src/app/g3w-ol/controls/geocodingcontrol.js
+++ b/src/app/g3w-ol/controls/geocodingcontrol.js
@@ -332,7 +332,10 @@ class Google {
     };
   };
   handleResponse(response= {}) {
-    const results = response.status === 'OK' ? response.results
+    let results;
+
+    if (response.status === 'OK') {
+      results = response.results
       .filter(result => ol.extent.containsXY(this.settings.extent, result.geometry.location.lng, result.geometry.location.lat))
       .map(result => {
         let name,
@@ -362,7 +365,12 @@ class Google {
             details: result.address
           }
         };
-      }) : [];
+      })
+    } else if (response.status === 'REQUEST_DENIED') { /* disable google geocoding provider on API key error */
+      this.active = false;
+      results = [];
+    }
+
     return {
       results,
       header: {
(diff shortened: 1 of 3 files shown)


volterra79:
@Raruto it could be a +1 if we can check if google key provide is enable of geocoding request

Raruto:
> it could be a +1 if we can check if google key provide is enable of geocoding request

@volterra79 checkout my latest commit: https://github.com/g3w-suite/g3w-client/pull/127/commits/b40fd026014e6b12190216e0c0e339b2614091ef

src/app/g3w-ol/controls/geocodingcontrol.js
```diff
@@ -575,16 +575,12 @@ function GeocodingControl(options={}) {
         const responses = await Promise.allSettled(promises);
         responses.forEach(({status, value: response}, index) => {
           if (status === 'fulfilled') {
-              const {header, results} = this.providers[index].handleResponse(response);
-              if (this.providers[index].active) {
```
Raruto: @volterra79  removing that condition i don't think it works like https://github.com/g3w-suite/g3w-client/pull/127/commits/b40fd026014e6b12190216e0c0e339b2614091ef anymore

src/app/g3w-ol/controls/geocodingcontrol.js
```diff
@@ -550,9 +549,10 @@ function GeocodingControl(options={}) {
         if (this.lastQuery === q && this.result.firstChild) { return; }
         const promises = [];
         /**
-         * For loop to provideres
+         * For loop to active Providers
          */
-        this.providers.forEach(provider => {
+        const activeProviders = this.providers.filter(provider => provider.active);
```
Raruto: moving that condition here is too early because a call to the google api has not yet been made to check that the key is valid (see: `Promise.allSettled(promises);` for more info)

Raruto (commented):
Check out my review

PR #130, opened 2022-08-09
Patch for a circular dependency found between `core/utils/geo.js` and `core/utils/ol.js`

### LINKED ISSUE #129: Bug when use Area or Length map control
## Subject of the issue
Use Length or Area map control we receice a javascript error (empty `Geometry` object).

## Environment
- g3w-client: **v3.5**

## Steps to reproduce

Screenshot from 2022-08-09 18-18-06

volterra79:
Close: https://github.com/g3w-suite/g3w-client/issues/129

### DIFF
--- a/src/app/core/utils/ol.js
+++ b/src/app/core/utils/ol.js
@@ -1,6 +1,9 @@
 // FIXME: remove weird import (utility functions should be stateles)
 import ApplicationState from "core/applicationstate";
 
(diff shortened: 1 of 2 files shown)


Raruto (approved):
In view of the ESM modules, we should try to use specialized utilities to try to solve this type of problem, for example:

- https://techsparx.com/nodejs/howto/circular-dependencies.html
- https://www.npmjs.com/package/madge

PR #132, opened 2022-08-10
Delete `src/app/g3w-ol/interactions/mouseEvents.js`

Raruto:
Remove the whole file as it contains only commented code


volterra79 (approved):
Test and works

PR #133, opened 2022-08-10
Delete `src/app/g3w-ol/utils/utils.js`

Raruto:
Remove the whole file as it is no longer needed after v3.5 (see: https://github.com/g3w-suite/g3w-client/pull/81)


volterra79 (approved):
Test and works

PR #94, opened 2022-07-26
Add `src/store` folder (registry singletons refactor)

Raruto:
Consolidate the following singletons definitions within a single folder:
  -  **`LayersStoresRegistry`**
  - **`ComponentsRegistry`**
  - **`ControlsRegistry`**
  - **`MapLayersStoresRegistry`**
  - **`PluginsRegistry`**
  - **`ProjectsRegistry`**
  - **`SessionsRegistry`** 

**List of changes:**

- create `src/store` folder
- deprecate notice for `core/catalog/cataloglayersstoresregistry.js`
- deprecate notice for `core/editing/sessionsregistryjs`
- deprecate notice for `core/map/maplayersstoresregistryjs`
- deprecate notice for `core/plugin/pluginsregistryjs`
- deprecate notice for `core/project/projectsregistryjs`
- deprecate notice for `gui/component/componentsregistryjs`
- deprecate notice for `gui/map/control/registryjs`

Closes: https://github.com/g3w-suite/g3w-client/issues/91

PR #134, opened 2022-08-10
Move `core/applicationstate` into `store/application-state`

Raruto:
- add `src/store/application-state.js`
- deprecate notice for `src/app/core/applicationstate.js`

Related to: https://github.com/g3w-suite/g3w-client/pull/94


volterra79:
@Raruto can you unify all import referred to applicationstate to core/applicationstate ?
ex. https://github.com/g3w-suite/g3w-client/blob/application-state/src/app/gui/map/vue/map.js#L1

volterra79 (approved):
Test and works

PR #140, opened 2022-08-12
Fix mapcontrols left alignement on mobile when keyboard is open

volterra79:
Remove add rh class on mapcontrols  on mobile because create a strange left alignement

## Before:

before

## After:

you can check that mapcontrols remain on left

Screenshot from 2022-08-12 11-24-28


volterra79:
@Raruto below you can test it and check that mapcontrols remain on left:

Screenshot from 2022-08-12 11-24-28

Raruto (approved):

PR #142, opened 2022-08-16
Update wrong project license within `package.json` (MPL-2.0)

Raruto:
**More info:**
- https://github.com/g3w-suite/g3w-client/blob/dev/COPYING.txt
- https://www.mozilla.org/en-US/MPL/2.0/


volterra79 (approved):
Done

PR #149, opened 2022-08-22
Fix issue on vue-template-compiler using named slot.

volterra79:
To fix issue on empty slot name value when parent component doesn't fill slot, at moment we need to use vue version 2.6.12 and same version for vue-template-compiler.

---

**NB** This pull also upgrade `"graceful-fs"` and package.json's `"engines"`.

Based on https://stackoverflow.com/questions/55921442/how-to-fix-referenceerror-primordials-is-not-defined-in-node-js :

**If you are using npm >= 8.3.0 in your package.json file:**

```
{
  // Your current package.json
  "overrides": {
    "graceful-fs": "^4.2.10"
  }
}
```
the solution above is the only on that solve the issue on npm install related to gulp@3

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,11 +35,11 @@
   "dependencies": {
     "regenerator-runtime": "^0.13.9",
     "shp-write": "^0.3.2",
-    "vue": "^2.6.12",
+    "vue": "2.6.12",
     "vue-cookie-law": "^1.13.3"
   },
-  "resolutions": {
-    "graceful-fs": "4.2.3"
+  "overrides": {
+    "graceful-fs": "^4.2.10"
   },
   "devDependencies": {
     "babel-core": "^6.26.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,11 +35,11 @@
...
     "regenerator-runtime": "^0.13.9",
     "shp-write": "^0.3.2",
-    "vue": "^2.6.12",
+    "vue": "2.6.12",
     "vue-cookie-law": "^1.13.3"
   },
-  "resolutions": {
-    "graceful-fs": "4.2.3"
+  "overrides": {
+    "graceful-fs": "^4.2.10"
```
Raruto, line 42: Please also add some notes about this change in the description of the pull request

package.json
```diff
@@ -118,13 +118,14 @@
...
     "vueify": "^9.4.1",
     "watchify": "^3.7.0",
     "yargs": "^8.0.2"
   },
   "browserslist": "defaults",
   "engines": {
-    "node": ">=16.13.2",
-    "npm": ">=8.1.2"
+    "node": ">=16.15.0",
+    "npm": ">=8.5.5"
```
Raruto, line 129: Please also add some notes about this change in the description of the pull request

Raruto (commented):
This pull also upgrade "graceful-fs" and package.json's "engines"

package.json
volterra79: Based on https://stackoverflow.com/questions/55921442/how-to-fix-referenceerror-primordials-is-not-defined-in-node-js :
**If you are using npm >= 8.3.0 in your package.json file:**

```
{
  // Your current package.json
  "overrides": {
    "graceful-fs": "^4.2.10"
  }
}
```
and after some test, the solution above is the only on that solve the issue on npm install related to gulp@3

Raruto (approved):

PR #155, opened 2022-08-25
Change default geocoding control marker icon (`yellow circle` --> `red marker`)

volterra79:
## List of changes:

- update default geocoding marker icon (`yellow circle` --> `red marker`)
- let marker icon stay visible on map until a new research is done (or the user clear results)

## Before

circle

## After

marker

Related to https://github.com/g3w-suite/g3w-client/pull/131


Raruto (commented):
@volterra79 please add a couple of pictures that shows the before / after so I get an idea of what to expect (from what you write it doesn't look like you're adding a new feature)

Raruto (approved):

PR #198, opened 2022-09-19
Fire `changeinput` event also on initial default value of input fields

### LINKED ISSUE #199: `changeinput` event not fired when a new feature field is set with a default value
### Checklist

### Subject of the issue

No trigger change event is trigger when new editing feature is created and one ore more field value is set with a default field config value.

### Steps to reproduce

1.  Create a new editing feature (editing config layer has a field with default value) and another field has an expression with this field
2.  No expression api is called

### Environment

All browsers

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
When create new editing feature and layer editing configuration layer has a default value to set when feature is created, can happen that if value of new feature is set with default value, not trigger change is fired. This missing trigger event can create a bug if another field of the feature is related whit an qgis field expression linked with the field.
This fix trigger change event input when a new feature is created and one or mode field have a default value

Closes #199

### DIFF
--- a/src/app/gui/inputs/service.js
+++ b/src/app/gui/inputs/service.js
@@ -5,7 +5,7 @@ function Service(options = {}) {
   // set state of input
   this.state = options.state || {};
   // in case is set default input value to input
-  this.set_input_default_value = false;
+  this.has_default_value = false;
   // type of input
   //this.state.validate.required && this.setValue(this.state.value);
   /*
@@ -44,7 +44,7 @@ proto.setValue = function(value) {
       }
     } else {
       this.state.value = this.state.input.options.default;
-      this.set_input_default_value = true;
+      this.has_default_value = true;
     }
   }
 };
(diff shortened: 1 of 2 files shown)


src/mixins/base-input.js
```diff
@@ -24,12 +24,15 @@ export default {
     },
     // called when input value change
     change() {
-      this.service.setEmpty();
+      this.service.setEmpty && this.service.setEmpty();
```
Raruto: @volterra79 this sounds weird, what are you trying to achieve exactly?

volterra79: @Raruto update

src/app/gui/inputs/service.js
```diff
@@ -4,6 +4,8 @@ const {t} = require('core/i18n/i18n.service');
 function Service(options = {}) {
   // set state of input
   this.state = options.state || {};
+  // in case is set default input value to input
+  this.set_input_default_value = false;
```
Raruto: @volterra79 is this a setter? maybe not, `has_default_value` or `has_default` would be more self-explanatory?

volterra79: @Raruto rename with has_default_value 7011a4a9b63a20eddd6cccd850742f490e213fad

Raruto (approved):

PR #209, opened 2022-09-23
Fix print area calculation on devices with `DEVICE_PIXEL_RATIO` greater than 1

### LINKED ISSUE #36: https://dev.g3wsuite.it/en/map/demo-34/ print Broken?
Hi, was just scrolling around the print part of https://dev.g3wsuite.it/en/map/demo-34/

and it seems the select of the map part in the print layout is broken as I cannot select the exact map zone of the print.

Is this a BUG or it is as intended?

Thank you!

image

image

volterra79:
Closes: #36 

Remove `ol.has.DEVICE_PIXEL_RATIO` from calculation of print area


tudorbarascu:
@volterra79 Thanks! Just tested it and it now works.

PR #221, opened 2022-09-29
Overall rename ambiguous variable name from `lng` to `language`

### LINKED ISSUE #182: Overall rename ambiguous property `lng`
As per **v3.6**, the `lng` shorthand can be easily confused with `longitude` (instead of `language`), for example:

### ApplicationState::lng

https://github.com/g3w-suite/g3w-client/blob/5789adafd2664a4c5603af257c37ef1b48fa5892/src/store/application-state.js#L14

https://github.com/g3w-suite/g3w-client/blob/8ff4f9f5ac84444d93a750b168987959c82957ed/src/app/gui/inputs/input.js#L19

https://github.com/g3w-suite/g3w-client/blob/6775405174fd293ab485d9450a03c50435e9afff/src/components/QueryResultsActionQueryPolygonCSVAttributes.vue#L40-L43

volterra79:
Closes: #182


src/store/application-state.js
```diff
@@ -11,7 +11,7 @@ const STATE = Vue.observable({
   download: false, // true if there is a downloaded that is waiting
   upload: false, // upload
   baseLayerId: null,
-  lng: 'en', // language default
+  language: 'en', // language default
```
Raruto, line 18: In order to avoid backwards compatibility issues we have also need to add here an alias between the old and the new property (ie. `ApplicationState.lng == ApplicationState.language` must be true at least until the next major release)

Raruto (changes requested):

src/store/application-state.js
volterra79: @Raruto ok

src/services/application.js
```diff
@@ -153,6 +153,7 @@ const ApplicationService = function() {
 
   this.changeLanguage = function(language){
     changeLanguage(language);
+    ApplicationState.lng = language; // DEPRECATED: this property will be removed on v4 version. Leave for backward compatibility
```
Raruto: Whenever possible make use of the [@deprecated](https://jsdoc.app/tags-deprecated.html) tag so that your IDE knows it too (syntax higliting, code linting, ..)

In the case of small annotations you can also use the inline version (to be used wisely...):

```js
/** @deprecated since v3.7 */
foo: "bar"
```

instead of:

```js
/**
 * @deprecated since v3.7
 */
foo: "bar"
```

Raruto (changes requested):

Raruto (approved):

PR #236, opened 2022-10-14
Replace the `docker compose` command with `docker-compose`

tudorbarascu:


tudorbarascu:
Actually, this is only to fix older docker compose commands. gonna close.

PR #243, opened 2022-10-24
Prevent to register event handlers for `undefined` setter methods (G3WObject)

volterra79:
In case we are trying to register:

- `foo.onbefore( 'setterName', function() { ... } )`
- `foo.onafter( 'setterName', function() { ... } )`

on a `setterName` method that is not present on `foo` object, we get the following javascript error:

```js
TypeError: Cannot reading properties of undefined (reading 'push')
at proto._onsetter (g3wobject.js:79:21)
at proto._onafter (g3wobject.js:27:15)
```

---
 
Screenshot from 2022-10-24 12-08-06

### DIFF
--- a/src/app/core/g3wobject.js
+++ b/src/app/core/g3wobject.js
@@ -73,20 +73,23 @@ proto.un = function(setter, key) {
   type=sync|async
 */
 proto._onsetter = function(when, setter, listener, async, priority=0, once=false) {
-  const settersListeners = this.settersListeners[when];
-  const listenerKey = `${Math.floor(Math.random()*1000000) + Date.now()}`;
+  let listenerKey;
+  const settersListeners = this.settersListeners[when] || {};
   const settersListeneres = settersListeners[setter];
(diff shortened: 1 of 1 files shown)


src/app/core/g3wobject.js
```diff
@@ -73,20 +73,23 @@ proto.un = function(setter, key) {
   type=sync|async
 */
 proto._onsetter = function(when, setter, listener, async, priority=0, once=false) {
-  const settersListeners = this.settersListeners[when];
-  const listenerKey = `${Math.floor(Math.random()*1000000) + Date.now()}`;
+  let listenerKey;
+  const settersListeners = this.settersListeners[when] || {};
   const settersListeneres = settersListeners[setter];
```
Raruto, line 78: @volterra79 `settersListeners` vs `settersListeneres` let's find out a different naming...

volterra79: @Raruto right. I'll refactor this part to be more readable

PR #247, opened 2022-10-26
Prevent auto-collapse of map theme selector after changing preset (catalog layers)

### LINKED ISSUE #246: Prevent auto-collapse of map theme selector after changing preset (catalog layers)
### Checklist

### Motivation

To avoid to reopen map theme section  when a map theme is open, it's would be better that map theme section stay open

### Suggested solution

After select a map theme, map theme section (color blue) stay open:

Screenshot from 2022-10-26 15-24-03

### Alternatives considered

_No response_

volterra79:
Closes: #246


leolami (approved):
OK for me

PR #248, opened 2022-10-27
Revert "Add new setter method `FeaturesStore::featuresLockedByOtherUser` (editing mode)"

volterra79:
Reverts g3w-suite/g3w-client#244


volterra79:
Mistake

PR #250, opened 2022-10-27
Add new function `src/app/core/utils/geo::createFeatureFromFeatureObject({ id, feature={} })` (copy and paste of feature attributes)

### LINKED ISSUE #9: Copy and paste from other layer
In the selection / confirmation session of the feature to be pasted, I would see more information.

immagine

For example, the list of all (or of the first three) fields of the associated record

Such as seen when merging multiple features (keeping the drop and the selection checkbox)

immagine

volterra79:
Closes: https://github.com/g3w-suite/g3w-client-plugin-editing/issues/9

### DIFF
--- a/src/app/core/layers/geolayermixin.js
+++ b/src/app/core/layers/geolayermixin.js
@@ -1,5 +1,6 @@
 const Projections = require('g3w-ol/projection/projections');
 const { getScaleFromResolution } = require('core/utils/ol');
+const { createFeatureFromFeatureObject } = require('core/utils/geo');
 const { sanitizeUrl } = require('core/utils/utils');
 const GUI = require('gui/gui');
 const RESERVERDPARAMETRS = {
@@ -98,7 +99,7 @@ proto.getOlSelectionFeatures = function(){
 
 proto.addOlSelectionFeature = function({id, feature}={}){
   this.olSelectionFeatures[id] = this.olSelectionFeatures[id] || {
-    feature,
+    feature: createFeatureFromFeatureObject({id, feature}),
     added: false
   };
   return this.olSelectionFeatures[id];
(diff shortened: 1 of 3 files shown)


src/app/core/layers/geolayermixin.js
```diff
@@ -64,17 +64,11 @@ proto.getOlSelectionFeature = function(id){
   return this.olSelectionFeatures[id];
 };
 
-proto.updateOlSelectionFeature = function({id, geometry}={}){
+proto.updateOlSelectionFeature = function({id, feature}={}){
```
Raruto, line 67: @volterra79 beware that this edit may not be backwards compatible, for example:

https://github.com/g3w-suite/g3w-client/blob/42781b4c12903419a07762563181652484ccc7b0/src/app/core/layers/tablelayer.js#L139

Is it also used by other plugins?

src/app/core/layers/geolayermixin.js
```diff
@@ -103,9 +97,9 @@ proto.getOlSelectionFeatures = function(){
   return this.olSelectionFeatures;
 };
 
-proto.addOlSelectionFeature = function({id, geometry}={}){
+proto.addOlSelectionFeature = function({id, feature}={}){
```
Raruto, line 100: @volterra79 the same goes for this other function, for example:

https://github.com/g3w-suite/g3w-client/blob/6e5b47a672ed6bd62cbc6f92f515c3effc1639c5/src/app/gui/table/tableservice.js#L160-L163

https://github.com/g3w-suite/g3w-client/blob/6e5b47a672ed6bd62cbc6f92f515c3effc1639c5/src/app/gui/table/tableservice.js#L253-L256

https://github.com/g3w-suite/g3w-client/blob/6e5b47a672ed6bd62cbc6f92f515c3effc1639c5/src/app/gui/table/tableservice.js#L160-L163

src/app/core/layers/geolayermixin.js
volterra79: It use just internally. Not plugin use it

src/app/core/layers/geolayermixin.js
volterra79: It used by queryresultservice.js when click on selection tool

PR #280, opened 2022-12-22
Use projection definition provide by server (including projection extent)

volterra79:
Related to https://github.com/g3w-suite/g3w-admin/pull/430

Add also a way to get Projection object definition using g3w-admin api `/crs/<EPSG_CODE>`

Example:

**Request:**

```log
/crs/3003
```

**Response:**

```json
{
    "result": true,
    "data": {
        "epsg": 3003,
        "proj4": "+proj=tmerc +lat_0=0 +lon_0=9 +k=0.9996 +x_0=1500000 +y_0=0 +ellps=intl +towgs84=-104.1,-49.1,-9.9,0.971,-2.917,0.714,-11.68 +units=m +no_defs",
        "geographic": false,
        "axisinverted": false,
        "extent": [
            1226046.6820902952,
            4047095.260762165,
            1727931.8998958569,
            5214000.012210975
        ]
    }
}
```


src/app/g3w-ol/projection/projection.js
```diff
@@ -1,15 +1,3 @@
...
- */
-const CUSTOM_PROJECTIONS_EXTENT = {
-  'EPSG:3876': [18835101.07,4367049.45,22702879.51,9383109.87],
-  'EPSG:32733': GRID_EXTENT_3857,
-  'EPSG:25833': [-2465144.8,3638055.41,2885759.28, 9493779.8]
-};
-
 const Projection = function(options={}) {
   if (!options.crs) return null;
   const {epsg, proj4:proj4def, geographic=false, axisinverted=false, extent} = options.crs; // new structure of information crs from server
```
Raruto: This type of comments (eg. without "external" references) are not very useful, if you really want to communicate something try to be more specific, otherwise it is just additional noise for the reader..

src/app/g3w-ol/projection/projection.js
```diff
@@ -19,7 +7,7 @@ const Projection = function(options={}) {
   ol.proj.Projection.call(this, {
     code: epsg,
     //FOR NOW FORCE TO GET EXTENT
-    extent: false ? extent : degrees ? GENERIC_GRID_EXTENT_DEGREE: CUSTOM_PROJECTIONS_EXTENT[epsg] || GENERIC_GRID_EXTENT,
```
Raruto, line 22:
```js
// FOR NOW FORCE TO GET EXTENT
```

Please avoid using comments that are valid only "for now" (yes but in what time?), or anything else that isn't exactly quantifiable..

Anyway, it seems that this comment is no longer needed (otherwise update it as you see fit)

src/app/gui/map/mapservice.js
```diff
@@ -1672,6 +1667,11 @@ proto._setupViewer = function(width, height) {
     this.moveDefaultLayersOnTop(zindex);
   });
 
+  /**
+   *
+   * Register remove addLayer
```
Raruto: improve documentation about it (it's quite ambiguous)

Raruto (changes requested):

src/app/g3w-ol/projection/projections.js
```diff
@@ -1,104 +1,55 @@
...
+    ol.proj.proj4.register(proj4);
     return projection;
   },
+
   /**
-   * extent get from https://epsg.io/
+   * Check and register epsg
+   * 
+   * @param epsg : "EPSG:<CODE>" Ex. "EPSG:4326"
+   * @returns {Promise<unknown>}
```
Raruto: Double check too, but I think the fulfilled type should be as follows:

```suggestion
   * @returns {Promise<Projection>}
```

Ref:

https://github.com/g3w-suite/g3w-client/blob/2affa0475d081d7f366dd5df3c33e8b18aac0d10/src/app/g3w-ol/projection/projections.js#L45

src/components/MapAddLayer.vue
```diff
@@ -81,6 +81,11 @@ import { Chrome as ChromeComponent } from 'vue-color';
 
 import { EPSG } from 'app/constant';
 
+/**
+ * @since v3.8
+ */
+const Projections = require('g3w-ol/projection/projections');
```
Raruto: The `@since` tag should only be used where you export something that is also accessible from outside that file (functions, variables, properties, classes, ...). If so, please double check and place the comment above the "exported" item / line, if not, just remove it

src/app/constant.js
```diff
@@ -349,6 +349,13 @@ export const ZINDEXES = {
   }
 };
 
+/**
+ * @since v3.8
+ */
+export const API_BASE_URLS = {
+  CRS: '/crs/'  //Example /crs/<epsg_code>
+};
+
```
Raruto, line 358: before making it public, did you check that this variable wasn't already useless?

Raruto (changes requested):

src/app/constant.js
volterra79: @Raruto the idea is to provide a unique point where all api urls from server are listed

PR #252, opened 2022-11-03
Remove deprecated files (from v3.6 / v.3.7) and replace them with related CJS to ESM imports

Raruto:
Permanently remove almost all files marked with a "deprecation notice" added during the **v3.6** / **v3.7** refactoring. Any changes made here should still be backward compatible with previous versions of g3w-client as they do not affect any published API.

## List of changes:

- Replaced by `src/app/constant.js`:
  - src/app/core/layers/filter/operators.js
  - src/app/gui/constraints.js
  - src/app/gui/print/printconfig.js
  - src/app/gui/search/constants.js

- Replaced by `src/mixins`:
  - src/app/gui/inputs/select/vue/selectmixin.js
  - src/app/gui/inputs/widgetmixins.js
  - src/app/gui/metadata/vue/components/metadatamixin.js

- Replaced by `src/app/core/utils/parsers`
  - src/app/core/parsers/parsers.js
  - src/app/core/parsers/response/parser.js
  - src/app/core/parsers/table/parser.js
  - src/app/core/parsers/vector/parser.js

- Replaced by `src/app/core/utils/validators`
  - src/app/core/validators/inputs/checkbox.js
  - src/app/core/validators/inputs/datetimepicker.js
  - src/app/core/validators/inputs/float.js
  - src/app/core/validators/inputs/integer.js
  - src/app/core/validators/inputs/radio.js
  - src/app/core/validators/inputs/range.js
  - src/app/core/validators/inputs/validator.js
  - src/app/core/validators/inputs/validators.js
  - src/app/core/validators/layers/validator.js

- Moved into `src/services`:
  - src/app/core/apiservice.js
  - src/app/core/clipboardservice.js
  - src/app/core/data/expression/service.js
  - src/app/core/data/ows/service.js
  - src/app/core/data/proxy/service.js
  - src/app/core/data/query/service.js
  - src/app/core/data/search/service.js
  - src/app/core/data/routerservice.js
  - src/app/core/editing/changesmanager.js
  - src/app/core/history_router.js
  - src/app/core/iframe/services/app/service.js
  - src/app/core/iframe/services/plugins/editing/service.js
  - src/app/core/iframe/services/plugins/index.js
  - src/app/core/relations/relationsservice.js
  - src/app/core/router.js
  - src/app/core/task/service.js
  - src/app/core/workflow/workflowsstack.js
  - src/app/gui/gui.js
  - src/app/gui/querybuilder/service.js

- Moved into `src/store`:
  - src/app/core/applicationstate.js
[...]

PR #288, opened 2023-01-05
Refactor and delete all file with comment "DEPRECATED: this file will be removed after v3.4"

volterra79:


Raruto:
@volterra79 I think this is quite similar to https://github.com/g3w-suite/g3w-client/pull/252

volterra79:
Duplicated https://github.com/g3w-suite/g3w-client/pull/252

PR #299, opened 2023-01-13
Update some file comments to better adhere to JSDoc guidelines

Raruto:
Double checked some code comments to ensure they use the following tags:

- [\@file](https://jsdoc.app/tags-file.html)
- [\@since](https://jsdoc.app/tags-since.html)
- [\@todo](https://jsdoc.app/tags-todo.html)
- \@fixme (although it's not part of the JSDoc standard, but it's quite a "de-facto".. 😟)

---

**For more info:**
- [JSDoc reference](https://jsdoc.app/)

### DIFF
--- a/src/app/api.js
+++ b/src/app/api.js
@@ -165,7 +165,7 @@ module.exports = {
       MapLayer,
       geometry: {
         Geometry,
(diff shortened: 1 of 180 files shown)


volterra79 (approved):
@Raruto very good job

PR #150, opened 2022-08-22
Replace `http-proxy` server with `BrowserSync` proxy (docker development)

Raruto:
BrowserSync already comes with a proxy server, so:

- make use of built in `browser-sync` proxy server within `gulpfile.js`
- ~~remove useless~~ deprecate variable `G3W_PROXY_ROUTES` within `config.template.js`
- remove gulp task `plugins`
- refactor gulp task `g3w-admin:plugins`
- refactor gulp task `select-plugins`

Refactor and simply vendors build chain:
- remove `src/assets/vendors/index.css.html`
- remove `src/assets/vendors/index.js.html`
- remove `src/index.dev.html`
- rename `src/index.prod.html` into `src/index.html`
- deprecate `config.template.js` variables (`distFolder`, `clientFolder`, `admin_static_folder`, `admin_templates_folder`)
- add new variable within `config.template.js` (`admin_overrides_folder`) and
- add new variable within `gulpfile.js` (`outputFolder`) in order to properly handle `dev` and `build` tasks

Handle local developed plugins:

- add new variable within `config.template.js` (`devConfig`)
- deprecate variable within `config.template.js` (`createProject`)
- deprecate variable within `config.template.js` (`setCurrentProject`)
- ~~automatically toggle `class="skin-blue"` on body tag (dev environment)~~
- refactor function `ApplicationService::obtainInitConfig()` and 
- move `require('app/dev')` into `src/app/main.js` (application entry point)
- remove mandatory `gid` and `baseurl` options within `G3W_PLUGINS`
  - convert variable `G3W_PLUGINS` into array (list of locally developed plugins) (ref. `config.template.js`)
  - automatically assing `gid` and `baseurl` parameter for all locally developed plugins (ref: `src/app/dev/index.js`)

**More info:**

- https://browsersync.io/docs/options#option-proxy

## How to test:

Update your `config.js` file as per [`516f734`](https://github.com/g3w-suite/g3w-client/blob/516f734bd3bdc48099ba2f6e532ee7ed34ca9ecc/config.template.js#L25-L26):

```js
 admin_plugins_folder:   '../g3w-admin/g3w-admin',                         // path to G3W-ADMIN where are stored all plugin folders 
 admin_overrides_folder: '../g3w-suite-docker/config/g3w-suite/overrides', // path to G3W-SUITE overrides folder 
```

Add the following lines at the top of your `g3w-suite-docker/config/g3w-suite/settings_docker.py` file

```py
#
[...]

Issue #63, opened 2022-07-13
Remove useful `gid` check within `initConfig` for current project while developing


Raruto:
In order to develop locally it is currently necessary to always change the value of the `gid` attribute for each plugin and project created:

https://github.com/g3w-suite/g3w-client/blob/42e51c463e58cf4eb1740f920c00de6c4bf6bcc7/src/config/dev/index.template.js#L20-L26

Related to: https://github.com/g3w-suite/g3w-client/pull/62

PR #301, opened 2023-01-13
Include default client plugins (editing, qplotly, qtimeseries) as git submodules

volterra79:
G3W-SUITE v3.5 already ships the following default plugins:

1. [editing](https://github.com/g3w-suite/g3w-client-plugin-editing)
2. [qplotly](https://github.com/g3w-suite/g3w-client-plugin-qplotly)
3. [qtimeseries](https://github.com/g3w-suite/g3w-client-plugin-qtimeseries)

Through the use of git submodules, this pull just makes sure that related source code is always present in the generated bundle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,14 +127,25 @@ http://localhost:3000/en/map/eleprofile/qdjango/2 # g3w-client (development)
 If you want develop custom plugins you need to place them in the [`src/plugins`](https://github.com/g3w-suite/g3w-client/blob/dev/src/plugins) folder, below you can see some examples:
 
 - [base-template](https://github.com/g3w-suite/g3w-client-plugin-base-template)
-- [editing](https://github.com/g3w-suite/g3w-client-plugin-editing)
 - [eleprofile](https://github.com/g3w-suite/g3w-client-plugin-elevation-profile)
 - [openrouteservice](https://github.com/g3w-suite/g3w-client-plugin-openrouteservice)
-- [qplotly](https://github.com/g3w-suite/g3w-client-plugin-qplotly)
-- [qtimeseries](https://github.com/g3w-suite/g3w-client-plugin-qtimeseries)
 - [queryresult-template](https://github.com/g3w-suite/g3w-client-plugin-queryresult-template)
 - [sidebar-template](https://github.com/g3w-suite/g3w-client-plugin-sidebar-template)
 
+### Default Plugins
+
+- [editing](https://github.com/g3w-suite/g3w-client-plugin-editing)
+- [qplotly](https://github.com/g3w-suite/g3w-client-plugin-qplotly)
+- [qtimeseries](https://github.com/g3w-suite/g3w-client-plugin-qtimeseries)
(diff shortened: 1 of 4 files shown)


config.template.js
```diff
@@ -6,9 +6,16 @@ const G3W_ADMIN_PORT   = '8000';      // G3W-ADMIN development server
...
-  // "your-plugin-folder-name-2",
-  // "your-plugin-folder-name-3",
+  // "<plugin-name>": {
+  //    <custom plugin configuration>
+  // }
+  /**
+   * Example:
+   * 'editing': {
+   *    baseurl: '../dist',
+   *    gid: 'qdjango:1'
```
Raruto: Check again, I thought this annoyance was already solved in here: https://github.com/g3w-suite/g3w-client/pull/150/commits/0e0d300b11f2d828c4508f2c0445881600887077

Ref: https://github.com/g3w-suite/g3w-client/issues/63

README.md
```diff
@@ -127,14 +127,25 @@ http://localhost:3000/en/map/eleprofile/qdjango/2 # g3w-client (development)
...
+- [editing](https://github.com/g3w-suite/g3w-client-plugin-editing)
+- [qplotly](https://github.com/g3w-suite/g3w-client-plugin-qplotly)
+- [qtimeseries](https://github.com/g3w-suite/g3w-client-plugin-qtimeseries)
+
+To install Default G3W-SUITE plugin run follow command:
+
+```sh
+
+git submodules init      # to initialize your local configuration file
+git submodule update     # to fetch all the data from that project and check out the appropriate commit listed in your project
```
Raruto, line 146: @volterra79 on which `git` version did you test the above commands?

volterra79: @Raruto git version 2.17.1

Raruto: I've read about different "default" behaviors depending on the installed version, actually mine are:

```log
git version 2.39.0.windows.2 # windows 10
git version 2.34.1           # windows subsystem for linux (ubuntu jammy)
```

Just to know, have you already updated your git version to the latest available on your OS?

volterra79: @Raruto yes. I'm working on Ubuntu 18.04.6 LTS

Raruto:
I feel like we're overcomplicating things here somehow.. (there would also be [git-subtrees](https://www.atlassian.com/git/tutorials/git-subtree) but I didn't go into details.)

At the moment, just to automate some steps, we could think of adding a npm task (`"preinstall"` or `"prebuild"`, I don't know) that takes care of cloning all the necessary plugins and keeping there all these folders as separate repositories.

Also because, if you think about it, to generate a coherent bundle for us it would be enough to be sure that all the default `src/plugins/**/plugin.js` files are there (and not the entire git repository), i.e.:

- `src/plugins/editing/plugin.js`
- `src/plugins/qplotly/plugin.js`
- `src/plugins/qtimeseries/plugin.js`

I'll think about it again in the next few days and I'll let you know.

Issue #98, opened 2022-07-28
List of all files and plugins that relys on `jQuery`


Raruto:
In upcoming releases we will only targeting more modern browsers, so we might not need anything more than what the browser ships with. 

Here are some tips to apply to migrate: [youmightnotneedjquery.com](https://youmightnotneedjquery.com/)

---

**TODO:** make a list of references to jquery below

---

Raruto:
As per **v4.1.0**, the last ones left to be removed:

- [select2](https://www.npmjs.com/package/select2)
- [eonasdan-bootstrap-datetimepicker](https://www.npmjs.com/package/eonasdan-bootstrap-datetimepicker)

PR #308, opened 2023-01-18
Migrate all jQuery promises to native ES6 `Promise` (ref: `$.Deferred()`, `$.when()`, `$.get()` and `$.post()`)

### LINKED ISSUE #82: Migrate all jQuery `$.Deferred` promises to native ES6 `Promise`
**Example 1 — Cleaning up a simple .then() and .fail()**

```js
// Before
function bar() {
 foo()
 .then(function() {
   console.log("Success");
 })
 .fail(function() {
   console.log("Error");
 });
}
```
```js
// After
async function bar() {
 try {
   await foo();
 console.log("Success");
 } catch {
   console.log("Error");
 }
}
```

**Example 2 — Cleaning up .then() and .fail() callbacks**

```js
// Before
function bar() {
 executeFoo(function() {
   console.log("Success");
 },
 function() {
   console.log("Error");
 });
}

function executeFoo(successCallback, errorCallback) {
 foo()
 .then(successCallback)
 .fail(errorCallback);
}
```
```js
// After
async function bar() {
 try {
   await foo();
   console.log("Success");
 } catch {
   console.log("Error");
 }
}
```

**Example 3 — Migrating $.Deferred() notifications**

```js
// Before
function notifyEvery500MS (progressCallback) {
 const deferred = $.Deferred();
 let count = 0; const interval = setInterval(function () {
   if (count === 10) {
     clearInterval(interval);
     deferred.resolve();
   }
   else {
     count += 1;
     deferred.notify(progressCallback);
   }
 }, 500);
 return deferred;
}
```
```js
// After (a bit more involved)
function notifyEvery500MS (progressCallback) {
 let count = 0; const interval = setInterval(function() {
   window.dispatchEvent(new CustomEvent("notify", {detail: count}));
   count += 1; 
 }, 500);return Promise.resolve({then: function(onfulfilled) {
   window.addEventListener("notify", function notifyListener(e) {
    if (e.detail === 10) {
       clearInterval(interval);
       window.removeEventListener("notify", notifyListener);
       onfulfilled();
    }
    else {
      progressCallback();
    }
   });
 }});
}
```

---

**Source:** https://medium.com/@szpytfire/migrating-from-jquery-to-es6-promises-65ee10553126

---

volterra79:
## List of changes

- replace `$.Deferred::done()` with `Promise::then()`
- replace `$.Deferred::fail()` with `Promise::catch()`
- replace `$.Deferred::alwas()` with`Promise::finally()`
- replace `$.when()` with `Promise::all()`
- replace `$.get()` with an internal utility function (ref: `src/core/utils/utils/XHR::get()`)
- replace `$.post()` with an internal utility function (ref: `src/core/utils/utils/XHR::post()`)
- ~~add [`"util-deprecate"`](https://www.npmjs.com/package/util-deprecate) dependency to effectively report deprecation of some features to final developer (ref: `src/deprecate.js`)~~ add new file `src/deprecate.js` to effectively report deprecation of some features to final developer
- add deprecation notice for `g3wsdk.core.utils.resolve(value)` function
- add deprecation notice for `g3wsdk.core.utils.reject(value)` function

Closes: https://github.com/g3w-suite/g3w-client/issues/82
Somehow related to: https://github.com/g3w-suite/g3w-client/issues/98


Raruto (changes requested):
Add a short description of the pull explaining to everyone which substitutions have been applied, for example:

- replace `$.Deferred::fail()` with `Promise::catch()`
- replace `$.Deferred::alwas()` with`Promise::finally()`
- replace `$.get() JQuery method with XHR.get that return a Promise ` 
- replace `$.post() JQuery method with XHR post() that return a Promise`
- replace `$.when() JQuery method with Promise.all() static method `

volterra79:
@Raruto i think that we can replace $.get, $.post with XHR.get and XHR.post for example here:

https://github.com/g3w-suite/g3w-client/blob/7b781924282e6571f62fd3661c4c0c42366de744/src/app/core/layers/providers/qgisprovider.js#L511

in order to use Promise.
Also $.when method can replaced with Promise static method (ex. Promise.all)
(based on https://api.jquery.com/jquery.when/)

https://github.com/g3w-suite/g3w-client/blob/7b781924282e6571f62fd3661c4c0c42366de744/src/app/core/layers/providers/qgisprovider.js#L383

Raruto:
> @Raruto i think that we can replace .get,.post with XHR.get and XHR.post for example here:
> 
> https://github.com/g3w-suite/g3w-client/blob/7b781924282e6571f62fd3661c4c0c42366de744/src/app/core/layers/providers/qgisprovider.js#L511
> 
> in order to use Promise. Also $.when method can replaced with Promise static method (ex. Promise.allSettled)
> 
> https://github.com/g3w-suite/g3w-client/blob/7b781924282e6571f62fd3661c4c0c42366de744/src/app/core/layers/providers/qgisprovider.js#L383

Ok, i think it's ok also because i don't have so deep knowledge of all these jquery methods..

Just remember that if you plan on going this way:

- always keep track of your main changes https://github.com/g3w-suite/g3w-client/pull/308#pullrequestreview-1261928202.
- (when applicable) update the title of the pull to make it more similar to the intent of these changes

In this way anyone can get an idea of any eventual shim functions to write to try to maintain backwards compatibility (eg. within `deprecated.js`), otherwise it is even more difficult to evaluate the changes made by reading each pull request file..

src/app/core/layers/map/vectorlayer.js
```diff
@@ -86,14 +86,14 @@ proto._makeStyle = function(styleConfig) {
...
-    })
-    .fail(err => d.reject(err));
-  return d.promise()
+  return new Promise((resolve, reject) =>{
+    this.provider.getFeatures(options)
+      .then(features => {
+        this.addFeatures(features);
+        resolve(features);
+      })
+      .catch(err => reject(err));
```
Raruto, line 93: double check, isn't this a superfluous statement?

```js
.catch(err => reject(err));
```

src/app/core/layers/providers/kmlprovider.js
```diff
@@ -11,9 +11,7 @@ inherit(KMLDataProvider, DataProvider);
 const proto = KMLDataProvider.prototype;
 
 proto.getData = function() {
-  const d = $.Deferred();
-  return d.promise();
+  return new Promise((resolve, reject) => {})
```
Raruto: double check, how can such a promise object be fulfilled?

```js
return new Promise((resolve, reject) => {})
```

volterra79: @Raruto i think that KMLDataProvider is never used. We need to double check and remove it

src/app/core/layers/providers/qgisprovider.js
```diff
@@ -455,7 +457,7 @@ proto._loadVectorData = function(vectorLayer, bbox) {
       vectorLayer.setData(vectorDataResponse.vector.data);
       if (this._) return vectorDataResponse;
     })
-    .fail(() => {
+    .catch(() => {
       return false;
     })
```
Raruto: double check, the following statement can be quite ambiguous:

```js
.catch(() => { return false; })
```

it is best to make explicit what the intent is (ie. `reject` or `resolve` the promise) by using one of the following statements:

```js
.catch(() => resolve(false))
```
```js
.catch(() => reject(false))
```

src/app/core/layers/providers/wfsprovider.js
```diff
@@ -12,137 +12,137 @@ inherit(WFSDataProvider, DataProvider);
...
-  $.post(url, params)
-    .then(response => d.resolve(response))
-    .fail(err => d.reject(err));
-  return d.promise();
+  return new Promise((resolve, reject) => {
+    XHR.post({
+      url,
+      data:params
+    }).then(response => resolve(response))
+      .catch(err => reject(err));
```
Raruto: double check, written that way wouldn't the whole surrounding promise above be superfluous?

```js
return // new Promise((resolve, reject) => {
    XHR.post({ ... })
//    .then(response => resolve(response))
//    .catch(err => reject(err));
// });
```

src/app/core/layers/providers/qgisprovider.js
volterra79: @Raruto you can replace .catch(() => { return false; }) with .catch(() => reject(false)). We double check if this method is still used.

src/app/core/layers/providers/wmsprovider.js
```diff
@@ -118,22 +118,22 @@ proto.query = function(options={}) {
...
-      d.resolve({
-        data,
-        query
-      });
-    })
-    .catch(err => d.reject(err))
-    .finally(()=> clearTimeout(timeoutKey));
-  return d.promise();
+      })
+    } else this[METHOD]({url, layers, params})
```
Raruto: 😩 Damn weird!

_else statement + missing brackets + a multiline istruction_

src/app/core/workflow/flow.js
```diff
@@ -16,21 +16,26 @@ function Flow() {
...
-    steps = workflow.getSteps();
-    // check if there are steps
-    if (steps && steps.length) {
-      //run step (first)
-      this.runStep(steps[0], inputs, context);
-    }
-    // return a promise that will be reolved if all step go right
-    return d.promise();
+    return new Promise((resolve, reject) => {
+      //** Assign d module variable to an object having resolve, reject methods used on other module method
```
Raruto: Specific code comments (jsdoc, but not only those) on the usefulness of a variable must be written when declaring it otherwise your development IDE is not able to recognize them, that is:

```js
/**
 * Module variable having resolve, reject methods used on other module method
 */
let d;

...

d = {
  resolve,
  reject
};
```

src/app/main.js
```diff
@@ -1164,7 +1164,7 @@ ApplicationTemplate.Services = {
   floatbar: null
 };
 
-ApplicationTemplate.fail = function({language='en', error }) {
+ApplicationTemplate.catch = function({language='en', error }) {
```
Raruto: **TODO:** also double check that it hasn't been renamed elsewhere

```suggestion
ApplicationTemplate.fail = function({language='en', error }) {
```

src/app/main.js
```diff
@@ -1205,7 +1205,7 @@ ApplicationService.init()
       if (error.responseJSON && error.responseJSON.error.data) error = error.responseJSON.error.data;
       else if (error.statusText) error = error.statusText;
     }
-    ApplicationTemplate.fail({
+    ApplicationTemplate.catch({
```
Raruto:
```suggestion
    ApplicationTemplate.fail({
```

src/services/data-query.js
```diff
@@ -135,7 +135,7 @@ function QueryService(){
       request.then(response => {
         const results = this.handleResponse(response, query);
         resolve(results);
-      }).fail(reject)
+      }).catch(reject)
```
Raruto, line 245: double check, written like that the external promise seems superfluous:

```js
return // new Promise((resolve, reject) => {
  request.then(response => {
    const results = this.handleResponse(response, query);
    resolve(results);
  })
// .catch(reject)
```

src/services/data-search.js
```diff
@@ -44,7 +44,7 @@ function SearchService(){
           resolve({
             data
           })
-        }).fail(reject)
+        }).catch(reject)
```
Raruto: same as: https://github.com/g3w-suite/g3w-client/pull/308/files#r1085451199

src/app/core/layers/map/vectorlayer.js
Raruto: > ```js
> .catch(err => reject(err));
> ```

I didn't mark them all, but I found this kind of repetition quite often in the code

src/app/core/layers/providers/kmlprovider.js
Raruto: ok, i didn't check in too much detail, but still from some other substitutions you've done i guess your initial intent was to write something like this instead:

```js
return Promise.resolve()
```

src/app/core/layers/map/vectorlayer.js
Raruto, line 93: Not sure what the previous behavior was, but now I think you need to add the return `(resolve, reject) => this.provider.getFeatures` from inside the promise to the outside to make `. catch(err => reject(err))` also "catchable" from outside of this function (otherwise it will only throw an unhandled fatal error):

```suggestion
return new Promise((resolve, reject) =>
    this.provider
      .getFeatures(options)
      .then(features => resolve(this.addFeatures(features)))
      .catch(err => reject(err))
);
```

Anyway i'm not so sure about this statement, check it carefully too

Raruto:
Superseded by: https://github.com/g3w-suite/g3w-client/pull/647

PR #310, opened 2023-01-20
Localize error page messages (ref: `Application::fail( { error } )`)

### LINKED ISSUE #45: Make the `ApplicationTemplate.fail({language='en', error })` function truly translatable
As you can see the following function allows the use of an alternative language, but in fact the values of i18n are hard coded inside the function itself instead of using a global configuration file:

https://github.com/g3w-suite/g3w-client/blob/6f844148168e844199e65254fb89173b4d55a3da/src/app/gui/app/index.js#L871-L884

I haven't checked if it's used elsewhere in the code, but for the moment I found this function within the main.js file

https://github.com/g3w-suite/g3w-client/blob/6f844148168e844199e65254fb89173b4d55a3da/src/app/main.js#L32-L35

Raruto:
## List of changes:

- delete file `src/app/gui/templates/500.html`
- add new i18n message `error_page.error`
- add new i18n message `error_page.at_moment`
- add new i18n message `error_page.f5`
- remove parameter `language` from  function `ApplicationTemplate.fail = function({ language='en', error })` (internally replaced with `t()` function)

Closes: https://github.com/g3w-suite/g3w-client/issues/45


src/locales/ro.js
```diff
@@ -1,5 +1,10 @@
 export default {
   translation: {
+    error_page: {
+      error: "Eroare de conexiune",
+      at_moment: "Momentan nu este posibil să afișați harta",
+      f5: "Apăsați Ctrl+F5"
+    },
```
Raruto, line 7: @tudorbarascu let me know if they need to be improved

Raruto:
@volterra79 can we merge this?

volterra79 (approved):

PR #311, opened 2023-01-20
Install `lodash.clonedeep` package

### LINKED ISSUE #57: Avoid direct usage of `JSON.parse(JSON.stringify(foo))`
It could be a bad practice to clone a js object in this way:

https://github.com/g3w-suite/g3w-client/blob/90e16348f9486158e0d65e403abcab25f60e9aa4/src/app/core/applicationservice.js#L631

https://github.com/g3w-suite/g3w-client/blob/6e5b47a672ed6bd62cbc6f92f515c3effc1639c5/src/app/core/layers/tablelayer.js#L447

Anyway, it would be better to introduce a utility function to declare the obscure sense of such a invocations chain.

Raruto:
## List of changes

- make explicit this dependency by installing the `lodash.clonedeep` package (ref: `package.json`)
- replace global references to window variable `_.cloneDeep` with internal `_cloneDeep` function (ref: `lodash.clonedeep`)
- ref: https://github.com/g3w-suite/g3w-client/issues/57
  - refactor function `ApplicationService::cloneLayout()` in order to make use of `lodash.clonedeep` package
  - refactor function `ApplicationService::cloneLayout()` in order to make use of `lodash.clonedeep` package

Closes: https://github.com/g3w-suite/g3w-client/issues/57


volterra79:
@Raruto we can take in accont https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

Raruto:
> we can take in accont https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

## [Features and limitations:](https://web.dev/structured-clone/#features-and-limitations)

Structured cloning addresses many (although not all) shortcomings of the `JSON.stringify()` technique. Structured cloning can handle cyclical data structures, support many built-in data types and is generally more robust and often faster.

However, it still has some limitations that may catch you off-guard:

- **Prototypes:** If you use `structuredClone()` with a class instance, you’ll get a plain object as the return value, as **structured cloning discards the object’s prototype chain.**
- **Functions:** If your object contains functions, they will be _quietly_ discarded.
- **Non-cloneables:** Some values are not structured cloneable, most notably `Error` and DOM nodes. It will cause `structuredClone()` to throw.

If any of these limitations are a deal-breaker for your use-case, **libraries like Lodash still provide custom implementations of other deep-cloning algorithms** that may or may not fit your use-case.

volterra79:
@Raruto i think we need to reduce dependencies instead to add one more. What do you think?

Raruto:
@volterra79 Deep cloning is a complex thing.

As is the browser's native `structuredClone` cannot replace the lodash's one (ref: https://github.com/g3w-suite/g3w-client/pull/311#issuecomment-1645048652).

In the future it would be possible to code some custom cloning algorithms, but for now it's simply best to make explicit the lodash dependency.

So let's start identifying the individual functions that will need to be replaced (that's why I didn't include the whole `lodash` library).

volterra79 (approved):

PR #316, opened 2023-02-01
Zoom to layer bounds after double clicking on related legend item (TOC)

volterra79:
In case of layer that has geometry, for example  in below image  if double click on **building**, it is possible to zoom to layer:

Screenshot from 2023-02-01 17-05-27

In case of single click,  **building** is select:

Screenshot from 2023-02-01 17-06-06

### DIFF
--- a/src/components/CatalogTristateTree.vue
+++ b/src/components/CatalogTristateTree.vue
@@ -5,7 +5,7 @@
 
 <template>
   <li
-    class="tree-item" @contextmenu.prevent.stop="showLayerMenu(layerstree, $event)" @click.prevent="select" :style="{marginLeft: !isGroup ? '5px' : '0'}"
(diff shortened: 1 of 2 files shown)


src/components/CatalogTristateTree.vue
```diff
@@ -81,6 +81,15 @@ import GUI from 'services/gui';
...
 
+/**
+ * @since v3.8
+ *
+ */
+const CLICKDOUBLECLICKEVENT = {
+  count: 0, // count click event
+  timeoutID: null // timeoutID return by setTimeout Function
+};
+
```
Raruto: 1. the `@since` tag is superfluous if the variable is not exported

2. explain better the use of the `CLICKDOUBLECLICKEVENT` variable, (eg. why is it not possible to use vue attributes? `@dblclick=` and `@click=`)

Raruto (changes requested):

Raruto (approved):

PR #328, opened 2023-02-10
Keep layer legend visible in a separate TAB regardless of layer scale dependency visibility

### LINKED ISSUE #327: Legend disappear when a layer change is visibility due scale dependence visibility
### Checklist

### Subject of the issue

If a layer change his visibility based on scale, legend on tab disappear

Screenshot from 2023-02-10 10-47-36

### Steps to reproduce

1. Set on QGIS project a layer with  **Scale dependency visibility**  (in our example **building**)

Screenshot from 2023-02-10 10-49-42

2. Save a QGIS project with layer at scale where layer set on point (1) is not visible

Screenshot from 2023-02-10 10-56-51

3. When Project is loaded on G3W-SUITE application, active legend tab

Screenshot from 2023-02-10 10-54-02

4. With legend tab active zoom to a scale where layer **building** layer is visible. Legend disappear

Screenshot from 2023-02-10 10-57-43

### Environment

- g3w-admin: v3.5
- g3w-client: v3.7.7
- browser: all

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes: #327 

Screenshot from 2023-02-10 11-01-36

### DIFF
--- a/src/components/CatalogLayersLegendItems.vue
+++ b/src/components/CatalogLayersLegendItems.vue
@@ -94,9 +94,8 @@ export default {
      * @returns {Promise<void>}
      */
     async getLegendSrc() {
-      /**
-       * need to be reset layers url
-       ***/
+
+      // reset layers url
       this.legendurls = [];
 
       await this.$nextTick();
--- a/src/components/CatalogLayersLegendItems.vue
+++ b/src/components/CatalogLayersLegendItems.vue
@@ -6,22 +6,21 @@
 <template>
   <div class="legend-item">
 
-    <figure v-for="legendurl in legendurls" :key="legendurl.url">
+    <figure v-for="legendurl in legendurls">
+        <bar-loader
+          :loading="legendurl.loading"
+        />
 
-      <bar-loader
-        :loading="legendurl.loading"
-      ></bar-loader>
+        <img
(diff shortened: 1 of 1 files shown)


src/components/CatalogLayersLegendItems.vue
```diff
@@ -91,22 +89,31 @@ export default {
       return catalogLayer && catalogLayer.getLegendUrl(params, { format: 'image/png', categories: layer.categories });
     },
 
-    async getLegendSrc(_layers) {
+    /**
+     * get legend src for visible layers
+     * @returns {undefined}
+     */
```
Raruto: Double check, I think the correct return would instead:

```suggestion
    /**
     * get legend src for visible layers
     * @returns {Promise<void>}
     */
```
it also seems that we are missing a `return` statement, otherwise how can the line following be executed?

```js
await this.getLegendSrc(this.layers); // it waits indefinitely?

changeLayersLegend.forEach(layer => layer.legend.change = false); 
```

**PS** perhaps we should found a more descriptive name for the `getLegendSrc()` function if it returns no value (ref: `@returns {undefined}`)

Raruto (approved):

PR #339, opened 2023-02-13
Ask for user geolocation only after clicking on related map control (ref: `geolocation` control)

### LINKED ISSUE #336: Disable or make automatic geolocation request a project-related option
### Checklist

### Motivation

Unless there is a real need when starting the application (e.g. the project it couldn't work without an explicit user's consent) the following prompt is a **very annoying anti-pattern**:

img

also because as you can see from the controls present in that project there is no geolocation button geolocation, so if a user wanted to geolocate himself at a later time he simply couldn't do it (and would remain somewhat confused, if not frustrated).

Ref: [map-controls](https://g3w-suite.readthedocs.io/en/latest/g3wsuite_client.html?highlight=geolocation#map-controls)

### Suggested solution

1. request the localization consent only if the geolocation button geolocation is activated in the current project
2. ask for localization consent at application startup only in the presence of real needs (e.g. weird `bounding box constraints`, a custom option in the `initiConfig`, etc.)

### Alternatives considered

_No response_

volterra79:
Closes: #336


volterra79:
@Raruto maybe it could be useful change error message to user to be more clear

Raruto:
Moved to milestone [v3.9](https://github.com/g3w-suite/g3w-client/milestone/9)

Raruto (approved):

PR #352, opened 2023-02-21
Check if `lightlayer !== null` within `LayersStore::createLayersTree(groupName, options={})`

### LINKED ISSUE #351: Uncaught TypeError: Cannot set properties of null (setting 'expanded')
### Checklist

### Subject of the issue

app.min.js: Uncaught TypeError: Cannot set properties of null (setting 'expanded')

### Steps to reproduce

When project config from server has a layer in layerstree that is not on layers property. 
Maybe is due a layer visibility setting on g3w-admin

### Environment

- g3w-admin: v3.5
- g3w-client: v.3.7.x
- browser: all

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes: #351

### DIFF
--- a/src/app/core/layers/layersstore.js
+++ b/src/app/core/layers/layersstore.js
@@ -292,27 +292,30 @@ proto.createLayersTree = function(groupName, options={}) {
     else {
       let traverse = (obj, newobj) => {
         obj.forEach(layer => {
-          let lightlayer = {};
-          if (layer.id !== null && layer.id !== undefined) {
-            if (tocLayersId.find(toclayerId => toclayerId === layer.id)) {
-              lightlayer.id = layer.id;
-            } else {
-              // is set to null in case of layer in layerstree
-              // but not in layers project configuration from server
-              lightlayer = null;
-            }
+          let lightlayer = null;
+          if (
+            null !== layer.id
+            && "undefined" !== typeof layer.id
+            && tocLayersId.find(toclayerId => toclayerId === layer.id)
+          ) {
+            // assign an object to lightlayer
+            lightlayer = {
+              id: layer.id
+            };
           }
           // case group
-          if (layer.nodes !== null && layer.nodes !== undefined) {
-            lightlayer.title = layer.name;
-            lightlayer.groupId = uniqueId();
-            lightlayer.nodes = [];
-            lightlayer.checked = layer.checked;
-            lightlayer.mutually_exclusive = layer["mutually-exclusive"];
+          if (null !==layer.nodes && "undefined" !== typeof layer.nodes) {
+            lightlayer = {
+              title: layer.name,
+              groupId: uniqueId(),
+              nodes: [],
+              checked: layer.checked,
+              mutually_exclusive: layer["mutually-exclusive"]
+            }; //assign an object
             traverse(layer.nodes, lightlayer.nodes)
           }
           // check if lightlayer is not null
-          if (lightlayer !== null) {
+          if (null !== lightlayer) {
             /** toggle expanded property for legend item (TOC) **/
             lightlayer.expanded = layer.expanded;
             newobj.push(lightlayer);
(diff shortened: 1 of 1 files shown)


Raruto:
Changed the name of this pull to something a little bit more technical, just for internal reference, because it's not so easy to understand which situation is being fixed (or if other bugs could be added through this change)

Raruto:
@volterra79 between these lines you can create a similar bug (ie. attempt to assign a property onto a null object):

https://github.com/g3w-suite/g3w-client/blob/742596c422202ac6f8477a31e462d3be309a4291/src/app/core/layers/layersstore.js#L299-L307

Check you too, but try to edit as follows to make it a bit more safer:

## • [src/app/core/layers/layersstore.js#L294-L313](https://github.com/g3w-suite/g3w-client/blob/8cb235f6caf335968b83bdc54d14fe9196735d31/src/app/core/layers/layersstore.js#L294-L313)

```js
obj.forEach(layer => {

  // set to null in case of layer in layerstree, but not in layers project configuration from server
  let lightlayer = null;            // EDIT: <-- set `null` instead of empty object `{}` as the initial value

  if (null !== layer.id && undefined !== layer.id && tocLayersId.find(toclayerId => toclayerId === layer.id)) {
    lightlayer = lightlayer || {};  // EDIT: <-- prevent assigning properties to a null object
    lightlayer.id = layer.id;
  }

  // case group
  if (null !== layer.nodes && undefined !== layer.nodes) {
    lightlayer = lightlayer || {}; // EDIT: <-- prevent assigning properties to a null object
    lightlayer.title = layer.name;
    lightlayer.groupId = uniqueId();
    lightlayer.nodes = [];
    lightlayer.checked = layer.checked;
    lightlayer.mutually_exclusive = layer["mutually-exclusive"];
    traverse(layer.nodes, lightlayer.nodes)
  }

  // check if lightlayer is not null
  if (null !== lightlayer) {
    lightlayer.expanded = layer.expanded; // toggle expanded property for legend item (TOC)
    newobj.push(lightlayer);
  }

});
```

src/app/core/layers/layersstore.js
```diff
@@ -288,28 +288,37 @@ proto.createLayersTree = function(groupName, options={}) {
...
-          if (layer.nodes !== null && layer.nodes !== undefined) {
-            lightlayer.title = layer.name;
-            lightlayer.groupId = uniqueId();
-            lightlayer.nodes = [];
-            lightlayer.checked = layer.checked;
-            lightlayer.mutually_exclusive = layer["mutually-exclusive"];
-            traverse(layer.nodes, lightlayer.nodes)
+          if (null !== layer.nodes && "undefined" !== typeof layer.nodes) {
+            lightlayer = ({
+              ...lightlayer,
```
Raruto, line 307: Feel free to revert my last commit https://github.com/g3w-suite/g3w-client/pull/352/commits/418ea78f16ebc759b3afadd98b9ea0a46436f79b if you think we'll never need keep a reference of the `id: layer.id` property within the second if as well (although that was the behavior before this pull)

src/app/core/layers/layersstore.js
```diff
@@ -288,38 +288,37 @@ proto.createLayersTree = function(groupName, options={}) {
...
-            && tocLayersId.find(toclayerId => toclayerId === layer.id)
-          ) {
-            // assign an object to lightlayer
-            lightlayer = {
-              id: layer.id
-            };
+
+          // case TOC has layer ID
+          if (null !== layer.id && "undefined" !== typeof layer.id && tocLayersId.find(id => id === layer.id)) {
+            lightlayer = ({ ...lightlayer, id: layer.id });
```
volterra79, line 301: @Raruto lightlayer is set null at beginning on every forEach cycle. I don't understand why use spread operator on it

Raruto (approved):
This pull seems syntactically correct.

For the bug fixed also refer to this other related one: https://github.com/g3w-suite/g3w-admin/pull/463

PR #367, opened 2023-03-03
Fallback to default widget value for `InputRange.vue` (optional + empty fields)

### LINKED ISSUE #366: Add default value to widget input range on editing form
### Checklist

### Motivation

In case of range widget input form that are not required 

Screenshot from 2023-03-03 11-55-54

if input value is cleared (set empty) need to be set to default value 

```js
// config input from server
{
    "name": "high",
    "type": "bigint",
    "label": "High",
    "editable": true,
    "validate": {},
    "pk": false,
    "default": "",
    "input": {
        "type": "range",
        "options": {
            "values": [
                {
                    "min": 10,
                    "max": 30,
                    "Step": 2,
                    "default": **null** // default value
                }
            ]
        }
    }
}
```

### Suggested solution

set input value to `input.options.values[0].default` in case of set empty

### Alternatives considered

_No response_

volterra79:
Closes: #366

### DIFF
--- a/src/components/InputRange.vue
+++ b/src/components/InputRange.vue
@@ -28,19 +28,32 @@ const Input = require('gui/inputs/input');
 export default {
   mixins: [Input],
   data() {
-    const options = this.state.input.options.values[0];
-    const min = 1*options.min;
-    const max = 1*options.max;
-    const step = 1*options.Step;
+    const {min, max, Step} = this.state.input.options.values[0];
     return {
-      max,
-      min,
-      step: step
+      min: 1*min,
+      max: 1*max,
+      step: 1*Step
     }
   },
   methods: {
     checkValue() {
-      const valid = this.state.validate.required || !_.isEmpty(_.trim(this.state.value)) ? this.service.getValidator().validate(this.state.value) : true;
+      // set initial valid to true
+      let valid = true;
+      // check if value of input is empty
+      const isEmpty = _.isEmpty(_.trim(this.state.value));
+      // in case of required input check if is not empty and check validity from validator
+      if (this.state.validate.required) {
+        valid =  !isEmpty && this.service.getValidator().validate(this.state.value);
+      } else {
+        //in case not required check if value is empty
+        if (isEmpty) {
+          //set default value
+          this.state.value = this.state.input.options.values[0].default;
+        } else {
+          // check validity from validator
+          valid = this.service.getValidator().validate(this.state.value);
+        }
+      }
       this.state.validate.valid = valid;
       this.change();
     }
--- a/src/components/InputRange.vue
+++ b/src/components/InputRange.vue
@@ -37,22 +37,19 @@ export default {
   },
   methods: {
     checkValue() {
-      // set initial valid to true
-      let valid = true;
+      // if state required initial value is false
+      let valid = !this.state.validate.required;
       // check if value of input is empty
       const isEmpty = _.isEmpty(_.trim(this.state.value));
       // in case of required input check if is not empty and check validity from validator
       if (this.state.validate.required) {
         valid =  !isEmpty && this.service.getValidator().validate(this.state.value);
+      } else if (isEmpty) {  //in case not required check if value is empty
+        //set default value
+        this.state.value = this.state.input.options.values[0].default;
       } else {
-        //in case not required check if value is empty
-        if (isEmpty) {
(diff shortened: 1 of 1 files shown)


src/components/InputRange.vue
```diff
@@ -28,19 +28,32 @@ const Input = require('gui/inputs/input');
...
+    const {min, max, Step} = this.state.input.options.values[0];
     return {
-      max,
-      min,
-      step: step
+      min: 1*min,
+      max: 1*max,
+      step: 1*Step
     }
   },
```
Raruto: This portion of code may be obscure, add a comment above explaining why we need to multiply by 1

src/components/InputRange.vue
```diff
@@ -28,19 +28,32 @@ const Input = require('gui/inputs/input');
...
+      } else {
+        //in case not required check if value is empty
+        if (isEmpty) {
+          //set default value
+          this.state.value = this.state.input.options.values[0].default;
+        } else {
+          // check validity from validator
+          valid = this.service.getValidator().validate(this.state.value);
+        }
+      }
```
Raruto: Not a big fan of this kind of solution:

1. if you're building a validator, it's always best to start with a false assumption

```js
// set initial valid to true
let valid = true;
```

2. if possible it is best to avoid nested else-if conditions, as they make the code more difficult to read and debug. Whenever possible consider writing simpler conditionals that remains at the same "indentation level" (it's also good for your brain..):

```suggestion
    checkValue() {

      // set initial valid to true
      let valid = true;

      // check if value of input is empty
      const isEmpty = _.isEmpty(_.trim(this.state.value));

      // required input, check if is not empty and check validity from validator
      if (this.state.validate.required) {
        valid = !isEmpty && this.service.getValidator().validate(this.state.value);
      }

      // not required input, check if value is empty and set default value
      else if (isEmpty) {
        this.state.value = this.state.input.options.values[0].default;
      }

       // check validity from validator
      else {
        valid = this.service.getValidator().validate(this.state.value);
      }

      this.state.validate.valid = valid;
      this.change();

    }
```

src/components/InputRange.vue
```diff
@@ -28,19 +28,29 @@ const Input = require('gui/inputs/input');
...
+        valid =  !isEmpty && this.service.getValidator().validate(this.state.value);
+      } else if (isEmpty) {  //in case not required check if value is empty
+        //set default value
+        this.state.value = this.state.input.options.values[0].default;
+      } else {
+        // check validity from validator
+        valid = this.service.getValidator().validate(this.state.value);
+      }
       this.state.validate.valid = valid;
       this.change();
```
Raruto: Check if written like this below could be equivalent:

```suggestion
      // check if value of input is empty
      const isEmpty = _.isEmpty(_.trim(this.state.value));
      // set default value in case not required check if value is empty
      if (!this.state.validate.required && isEmpty) {
        this.state.value = this.state.input.options.values[0].default;
      }
      // TODO: add description
      this.state.validate.valid = this.state.validate.required && !isEmpty && this.service.getValidator().validate(this.state.value);
      this.change();
```

volterra79: @Raruto no. Because:
in case of `this.state.validate.required === false` has to be `this.state.validate.valid = true` if `isEmpty`, othewise  need to set `valid` value according of `this.service.getValidator().validate(this.state.value)` response

Raruto: Ok, let's try to simplify it one step at a time, because there seems to me something strange or wrong in the overall logic (ref: that double call to the `this.service.getValidator().validate(this.state.value)` function).

So, currently, is it equivalent to the following?

```js
checkValue() {
  // check if value of input is empty
  const isEmpty = _.isEmpty(_.trim(this.state.value));

  // in case not required check if value is empty and set default value
  if (isEmpty && !this.state.validate.required) { 
    this.state.value = this.state.input.options.values[0].default;
  }

  // if state required initial value is false
  this.state.validate.valid = !this.state.validate.required;

  // if is not empty check validity from validator
  if (!isEmpty) {
    this.state.validate.valid = this.service.getValidator().validate(this.state.value);
  }

  this.change();
}
```

volterra79: @Raruto done

Raruto (approved):
Ok, I'd say the syntax is fine for now

PR #376, opened 2023-03-13
Add support for multi layers search (ref: `selectBox` and `autocompleteBox` widgets)

### LINKED ISSUE #375: Show list of values on each search fields (select/autocomplete box widget) coming from main layer an other layers
### Checklist

### Motivation

In case of MultiLayer search widge, it could be ncessary have for each field list on serach form, in case selectBox widget, all values coming from main layer and other layers

Screenshot from 2023-03-13 15-21-19

### Suggested solution

For each layer (main layer and Other searching layers) , if widget input is a select/autocomplete box, list values coming from the field name of each layer involved and not only in main layer

### Alternatives considered

_No response_

volterra79:
Closes: #375


src/app/gui/search/vue/panel/searchservice.js
Raruto, line 1: It could be be written in a more general way? (eg. make use of only a single function like `getLayersFilterData`, regardless of the number of search layers used?)

Below a draft:


```js
/**
 * Get unique values from field
 * 
 * @param {{ field, value, unique }} options
 * 
 * @returns {Promise<[]>}
 */
proto.getUniqueValuesFromField = async function({field, value, unique}) {
  let data = [];
  try {
    await this.getLayersFilterData(
      // layers
      (
        1 === this.searchLayers.length
          ? [this.searchLayer] // In case of single layer to search
          : this.searchLayers  // In case other then single layer are set Other searching layers
      ),
      // options
      {
        suggest: (undefined !== value) ? `${field}|${value}` : undefined,
        unique,
        ordering: field.attribute
      }
    );
  } catch(err) {
    console.warn(err);
  }
  return data;
};

/**
 * @param {{field, value}} [options = {}]
 * 
 * @returns {Array}
 */
proto.autocompleteRequest = async function({field, value} = {}) {
  let data = [];
  try {
    data = await this.getLayersFilterData(
      (
        1 === this.searchLayers.length
          ? [this.searchLayer] // In case of single layer to search
          : this.searchLayers  // In case other then single layer are set Other searching layers
      ),
      {
        field: this.getAutoFieldDependeciesParamField(field),
        suggest: `${field}|${value}`,
        unique: field
      }
    );
  } catch(err) {
    console.warn(err)
  }
  return data.map(value => ({ id:value, text:value }));
};
```

volterra79: @Raruto let me check

Raruto: @volterra79 so in the end, did you find out something like the above https://github.com/g3w-suite/g3w-client/pull/376#discussion_r1135132692?

volterra79: @Raruto not yet. I'm working on it

Raruto:
@volterra79 Have you already checked that it works with my latest changes?

It might take a while, but I can do more in-depth tests if you need.

src/app/gui/search/vue/panel/searchservice.js
volterra79: @Raruto we leave as it is for now

volterra79:
> @volterra79 Have you already checked that it works with my latest changes?
> 
> It might take a while, but I can do more in-depth tests if you need.
@Raruto i'll check

Raruto (approved):

PR #320, opened 2023-02-07
Deprecate hooks `ProjectsRegistry~setCurrentProject(project)` and `ProjectsRegistry~createProject(projectConfig)`

Raruto:
Start addressing: https://github.com/g3w-suite/g3w-client/issues/67

---

As a result of: https://github.com/g3w-suite/g3w-client/pull/150 the intent is to bring the development environment ever closer to the production one. Consequently in a near future these lines of code will no longer be useful for local development:

https://github.com/g3w-suite/g3w-client/blob/69fcc81fa695782bda5d75ff05b36e9b59d41f95/config.template.js#L38-L41

Related to: https://github.com/g3w-suite/g3w-client/issues/89, https://github.com/g3w-suite/g3w-client/issues/292 and https://github.com/g3w-suite/g3w-suite-docker/pull/79

PR #386, opened 2023-03-21
Remove promise return from function `ApplicationService::changeProject({ gid, host })`

volterra79:
No more return promise is used when changeProject is run because  location.replace(url) reload page

### DIFF
--- a/src/components/ProjectsMenu.vue
+++ b/src/components/ProjectsMenu.vue
@@ -51,21 +51,7 @@ export default {
         const {gid} = item;
         item.cbk.call(item, {
           gid
-        }).then(promise => {
-            //changeProject is a setter so it return a promise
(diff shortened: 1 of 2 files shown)


Raruto:
Moved to [v3.9](https://github.com/g3w-suite/g3w-client/milestone/9) because I have the suspect that calling `location.replace()` the context change isn't so immediate and removing the promised return from the `changeProject()` function might break something else (ie. anything like `changeProject().then(...)`).

Remember to check better when we get closer to:

- https://github.com/g3w-suite/g3w-client/pull/320
- https://github.com/g3w-suite/g3w-client/pull/308

Raruto:
Superseded by: https://github.com/g3w-suite/g3w-client/pull/828

PR #390, opened 2023-03-23
Allow to print external WMS layers with same origin URL of application (ref: `"geoscreenshot"` and `"screenshot"` map controls)

volterra79:
Screenshot from 2023-04-13 11-23-26

Geoscreenshot and Screenhot map control are not add to map controls in case of layer with source **external true**:

```
{
  "type": "wms",
  ....
  "styles": "",
  "url":........,
  "external": true
}
```

because CORS issue occurs.

It can happen that url of external source has that same domain of the current application. In this case it add an additional check in case of source external true comparind doamin of current application with url of external source. If no CORS issue is present, map controls Geoscreenshot and Screenshot are added


src/app/g3w-ol/controls/screenshotcontrol.js
```diff
@@ -20,9 +20,27 @@ proto.change = function(layers=[]){
...
+       * map image
+       */
+      try {
+        const domain = new URL(layer.getSource().url);
+        // check if same origin of current page
+        return location.origin !== domain.origin;
+      } catch(err){}
+      return true;
+    } else return false
+  });
```
Raruto: I haven't tested, but to make it a little more immediate to the reader (and reusable elsewhere) I'd recommend rewriting it into something like the following:

```suggestion
/**
 * Check visibility map control based on layers
 *
 * @param {array} layers
 * @returns {boolean}
 */
proto.checkVisible = function(layers=[]) {
  return "undefined" === typeof (layers.find((layer) => sameOrigin(location.href, layer.getSource().url)));
};

/**
 * @TODO move this function into a shared "utility" file
 * @TODO handle try..catch for malformed URLs
 * 
 * Check for cross origin URLs.
 *
 * @since 3.8.0
 */
function sameOrigin(a, b) {
  return (new URL(a)).origin === (new URL(b)).origin;
}
```

volterra79: @Raruto we need to check is layer has source url otherwise sameOrigin method return an error because new URL(undefined) in not valid

Raruto:
> It can happen that url of external source has that same domain of the current application. In this case it add an additional check in case of source external true comparind doamin of current application with url of external source. If no CORS issue is present, map controls Geoscreenshot and Screenshot are added

@volterra79 can you provide same sample URLs? (just to figure out when we are sure that there's no CORS error)

volterra79:
@Raruto 

```json
{
    "type": "wms",
    "crs": "EPSG:3857",
    "dpiMode": "7",
    "format": "image/png",
    "layers": "rt_ofc.5k16.32bit.1R4G2B",
    "styles": "",
    "url": "http://www502.regione.toscana.it/ows_ofc/com.rt.wms.RTmap/wms?map=owsofc",
    "external": true
}

```

Raruto (approved):
> ```json
> {
>    "type": "wms",
>    "crs": "EPSG:3857",
>    "dpiMode": "7",
>    "format": "image/png",
>    "layers": "rt_ofc.5k16.32bit.1R4G2B",
>    "styles": "",
>    "url": "http://www502.regione.toscana.it/ows_ofc/com.rt.wms.RTmap/wms?map=owsofc",
>    "external": true
> }
> ```

I don't quite understand what are the two URLs to test with, but it seems good to me

PR #392, opened 2023-03-24
Add shortcut links within catalog layers context menu (layers and project settings)

### LINKED ISSUE #391: Link to the edit project page
### Checklist

- [x] I've searched through the [current issues](https://github.com/g3w-suite/g3w-client/issues) to make sure this feature hasn't been requested already.

### Motivation

As per **v3.8.0.alpha.2**,

I don't think there are any other references to the admin interface within the client application.

Let's try to facilitate at the administration of a single project in order to be able to reach the relative management page more easily (ie. without so many clicks ..).

### Suggested solution

For example, we might think about adding a link in here:

image

to be redirected to the following page:

image

### Alternatives considered

The group management page might also deserve a link within the client application, but we're not that greedy.. 😋

image

volterra79:
Closes: #391

### 1. General project settings 👉 [options and actions](https://g3w-suite.readthedocs.io/en/v3.6.x/g3wsuite_administration.html#options-and-actions)

image

### 2. Project layers settings 👉 [widget management](https://g3w-suite.readthedocs.io/en/v3.6.x/g3wsuite_administration.html#widget-management)

image

## How to test

Save the project again or disable project caching: `QDJANGO_PRJ_CACHE = False`


src/components/Catalog.vue
```diff
@@ -7,73 +7,141 @@
   <!-- item template -->
   <div id="catalog" @contextmenu.prevent.stop="" class="tabbable-panel catalog">
     <div class="tabbable-line">
+
+      <!-- TAB MENU (header) -->
       <ul class="nav nav-tabs catalalog-nav-tabs" role="tablist" @click.capture="delegationClickEventTab">
-        <li v-if="hasLayers" role="presentation"  :class="{ active: activeTab === 'layers' && 'hasLayers' }">
+        <li v-if="hasLayers" role="presentation"  :class="{ active: ('layers' === activeTab  && 'hasLayers') }">
```
Raruto:
```js
&& 'hasLayers'
```

Isn't this statement always true? (ie. are single quotes somehow needed?)

volterra79: @Raruto there is a case (rare case) that project has only base layers

Raruto: > there is a case (rare case) that project has only base layers

Ok, but my question is much more trivial

That means, what's the actual difference between these two predicates?

```js
&& hasLayers // <-- this should be a computed variable or method
```

```js
&& 'hasLayers' // <-- is this just a string? (I've never seen it before..)
```

volterra79: @Raruto i think is wrong. Can you delete it. It is a string and it is always true

Raruto:
## General project settings 👉 [options and actions](https://g3w-suite.readthedocs.io/en/v3.6.x/g3wsuite_administration.html#options-and-actions)

image

## Project layers settings 👉 [widget management](https://g3w-suite.readthedocs.io/en/v3.6.x/g3wsuite_administration.html#widget-management)

image

Raruto:
@volterra79 we could open those "layers/project settings links into a modal overlay (iframe), instead of a new browser tab.

Do you think that is something we can easily do with already available components?

Raruto:
@volterra79 can you check latest conflicts on this and then merge?

Raruto:
@volterra79 is it okay for you, can we merge it into `dev`?

Raruto (approved):

PR #417, opened 2023-05-18
Immediately remove query results layer from map (marker, bbox, etc..) when related control is untoggled

### LINKED ISSUE #416: QueryByDrawPolygon, polygon drawed not disappeared
### Checklist

### Subject of the issue

After drawing a polygon for to execute a search on a map area, if I try to change tools (i.e. switch to other query system) the polygon that I drew not earlier doesn't disappear:

Screenshot_20230518_144328

Screenshot_20230518_144344

### Steps to reproduce

1. active `QueryByDrawPolygon`
2. draw a polygon
3. activate another query tool.

### Environment

- g3w-admin: v3.6.0
- g3w-client: 3.8.0
- browser: Firefox
- operating system: Ubuntu 22.04 desktop

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes: #416

### DIFF
--- a/src/app/g3w-ol/controls/basequerypolygoncontrol.js
+++ b/src/app/g3w-ol/controls/basequerypolygoncontrol.js
@@ -29,7 +29,11 @@ const BaseQueryPolygonControl = function(options = {}) {
       type: 'spatialMethod',
       how: 'toggled' // or hover
     },
-    onhover: true
+    onhover: true,
+    /**
+     * @since 3.8.1
+     */
+    queryResultsLayer: true,
   };
 
   options = merge(options, default_options);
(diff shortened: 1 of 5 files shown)


Raruto:
I would recommend keeping the right sidebar panel open if the user switches to another query tool.

It can get a little confusing, with all those horizontal layout changes so close together:

demo

volterra79:
> I would recommend keeping the right sidebar panel open if the user switches to another query tool.
> 
> It can get a little confusing, with all those horizontal layout changes so close together:
> 
> demo

@wlorenzetti @leolami what you think about?

PR #448, opened 2023-07-13
Adjust tooltip color of `Enlarge / Reduce` icon + wrong `Default size` on vertical viewports (ref: `components/GlobalResizeIcon.vue`)

### LINKED ISSUE #447: Wrong tooltip background color on enlange/reduce icon
### Checklist

### Subject of the issue

When enlange/reduce icon is relative to verical side, toolip backgrond color is not related to skin color but is set as default (black) 

### Steps to reproduce

1. Do  a query

Screenshot from 2023-07-13 17-00-22

3.Open attribute table of a layer 

Screenshot from 2023-07-13 17-02-21

### Environment

- g3w-admin: 3.6.x
- g3w-client: 3.8.x
- browser: all

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes: #447 

Screenshot from 2023-07-13 17-07-23

### DIFF
--- a/src/services/application.js
+++ b/src/services/application.js
@@ -752,39 +752,34 @@ const ApplicationService = function() {
     return d.promise();
   };
 
+  /**
+   * Updates panels sizes when showing content (eg. bottom "Attribute Table" panel, right "Query Results" table)
+   */
   this.setLayout = function(who='app', config={}) {
-    /**
-     * Set default height percentage of height when show
-     * vertical content (for example show table attribute)
-     */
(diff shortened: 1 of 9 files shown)


src/services/application.js
Raruto, line 1: a bit weird logic, but for now let's keep it that way..

volterra79: @Raruto i agree.  We think about it on next releases

Raruto (approved):

PR #454, opened 2023-07-21
Delete folder `src/app/core/expression/` (internal helpers refactor)

Raruto:
## List of changes

Consolidate all helper the functions within `src/app/core/expression/` folder into the only two places where they are currently used:

- move `src/app/core/expression/inputservice.js` into `src/app/gui/form/formservice.js`
- move `src/app/core/expression/tabservice.js` into `src/components/GlobalTabs.vue`


volterra79:
@Raruto do you think that we can move expression `inputservice.js` and `tabservice.js` inside `services` folder instead move code inside `formservice.js` and `GlobalTabs.vue` component? In this way we can remove `expression` folder as well

Raruto:
@volterra79 currently the `src/services` folder contains singleton objects, eg. the `FormService` class is not:

https://github.com/g3w-suite/g3w-client/blob/795b2138715ed65bb0120ce212ba48fbdf127a3b/src/app/gui/form/formservice.js#L6

https://github.com/g3w-suite/g3w-client/blob/795b2138715ed65bb0120ce212ba48fbdf127a3b/src/app/gui/form/formservice.js#L666

So in general, leave code definition as-is if it's not global variable.

**NB** `tabservice.js` would be `utils` function instead, but since it is only used once it is better not to expose it elsewhere:

https://github.com/g3w-suite/g3w-client/blob/795b2138715ed65bb0120ce212ba48fbdf127a3b/src/app/core/expression/tabservice.js#L6-L18

👋 Raruto

volterra79 (approved):

PR #483, opened 2023-09-06
Add support to editable vector joins (1:1 relations)

volterra79:
Related to:

- https://github.com/g3w-suite/g3w-admin/pull/599
- https://github.com/g3w-suite/g3w-client-plugin-editing/pull/52

## • father_join

image

image

## • child_join

image

image

## QGIS configuration --> [16.1.10. Joins Properties](https://docs.qgis.org/3.28/en/docs/user_manual/working_with_vector/vector_properties.html#joins-properties)

image


src/app/constant.js
```diff
@@ -10,6 +10,14 @@
...
 import version from '../version';
 
+/**
+ * @since v3.7.0
+ */
+
+import eventbus from "./eventbus";
+
+export const APP_EVENTBUS = eventbus;
+
```
Raruto: Re-evaluate the implications

volterra79: @Raruto need it to use a central event bus application used by plugins

Raruto: @volterra79 Let me know if doing like so could be okay too 👉 https://github.com/g3w-suite/g3w-client-plugin-editing/pull/52/commits/178a3fd9cae1fa77d44050feda3684aa7fbf34af + https://github.com/g3w-suite/g3w-client/pull/483/commits/6818ac30ed2ee43b89a7088a548a05f7cdb86526 

Unless strictly necessary, I would prefer avoid exposing new global stuff.

Raruto (approved):

src/app/core/editing/editor.js
```diff
@@ -55,6 +55,10 @@ function Editor(options = {}) {
    */
   this._featuresstore = this._layer.getType() === Layer.LayerTypes.TABLE ? new FeaturesStore() : new OlFeaturesStore();
 
+  if (this._layer.getType() === Layer.LayerTypes.TABLE) {
+    PIPPO = this._featuresstore;
+  }
+
```
Raruto: @volterra79 double check

PR #517, opened 2023-11-07
Fix no results found for 1:N relations (search widget)

volterra79:
## How to test

1. Create a search on 1:n relation

Screenshot from 2023-11-07 12-00-07

2. Do a search that would return more than one result:

## Before

Screenshot from 2023-11-07 11-50-10

## After

Screenshot from 2023-11-07 11-48-06


Raruto (approved):
seems fine

PR #493, opened 2023-10-05
Remove API `g3wsdk.core.editing` related to `g3w-client-plugin-editing` repository

Raruto:
Related to: https://github.com/g3w-suite/g3w-client-plugin-editing/pull/65

## Motivation

1. Move most of the [g3w-client-plugin-editing](https://github.com/g3w-suite/g3w-client-plugin-editing)'s code to its dedicated repository and wait here for its API to be globally available, eg:

```js
# src/app/core/layers/tablelayer.js#L174

await _waitFor(() => window.g3wsdk.core.hasOwnProperty('editing')); // wait unitil "editing" plugin is loaded
```
**NB** all others editing-dependent plugins we have are already waiting for it to be well established before starting themselves (so it's no a real API breaking change):

```js
  this.getDependencyPluginsObject(['editing']).then(..)
```

2. reduce the overall number of editing related PRs

## List of changes

- delete folder: `src/app/core/editing/`
- delete file: `src/services/editing.js`
- delete file: `src/store/sessions.js`
- remove [client](https://github.com/g3w-suite/g3w-client) API: `g3wsdk.core.editing` (will be replaced by [editing](https://github.com/g3w-suite/g3w-client-plugin-editing) API)

PR #526, opened 2023-12-13
Handle deep relations on editing [g3w-admin#PR685]

volterra79:
## Related to: https://github.com/g3w-suite/g3w-client-plugin-editing/pull/80

Pass deep relations on commit items object

Screenshot from 2024-01-09 11-14-39


src/app/core/editing/session.js
```diff
@@ -322,44 +406,91 @@ proto._serializeCommit = function(itemsToCommit) {
...
+        //add to deleteRelationIds
+        deleteRelationIds.push(rlId);
+      }
+    })
+
+    //delete deep relations from current layer relation object
+    //because are not relative of this layer
+    deleteRelationIds.forEach(rId => delete commitObj.relations[rId]);
+  }
+
```
Raruto: @volterra79 I'm quite convinced it would be better to implement above changes after this other ones:

- https://github.com/g3w-suite/g3w-client-plugin-editing/pull/65
- https://github.com/g3w-suite/g3w-client/pull/493

Additionally, all editing-dependent plugins we have are already waiting for it to be well established before starting themselves (so it's no a real API breaking change):

```
this.getDependencyPluginsObject(['editing']).then(..)
```

Please take a closer look you too, so we can start thinking decreasing all those (duplicated) editing related PRs..

volterra79: @Raruto i think is not possible to wait  because it need to pay attention and some test to check if plugins related to editing plugin works properly. Changes related to deep relation management need to be ready next week for g3w-suite v3.7.2.
This pull request is related to https://github.com/g3w-suite/g3w-client-plugin-editing/pull/80

src/app/core/editing/session.js
```diff
@@ -322,44 +406,91 @@ proto._serializeCommit = function(itemsToCommit) {
...
+        commitObj.relations[relationId].relations[rlId] = commitObj.relations[rlId];
+        //add to deleteRelationIds
+        deleteRelationIds.push(rlId);
+      }
+    })
+
+    //delete deep relations from current layer relation object
+    //because are not relative of this layer
+    deleteRelationIds.forEach(rId => delete commitObj.relations[rId]);
+  }
```
Raruto:
```suggestion
  // Remove deep relations from current layer (commitObj) that are not relative to that layer
  let relations = Object.keys(commitObj.relations || []);
  relations
    .filter(id => undefined === this._editor.getLayer().getRelations().getArray().find(r => id === r.getChild())) // child relations
    .map(id => {
      commitObj.relations[
        SessionsRegistry
          .getSession(id)
          .getEditor()
          .getLayer()
          .getRelations()
          .getArray()
          .find(r => -1 !== relations.indexOf(r.getFather())) // parent relation layer
          .getFather()
        ].relations[id] = commitObj.relations[id];
      return id;
    })
    .forEach(id => delete commitObj.relations[id]);
```

Raruto (approved):

PR #534, opened 2023-12-27
Add support for `chain filter` option within relation reference widget [g3w-client-plugin-editing#63]

volterra79:
Related to https://github.com/g3w-suite/g3w-client-plugin-editing/pull/83


Raruto (approved):
Tested with: [editing-chain-filter-relation.qgs](https://github.com/g3w-suite/g3w-suite-demo-projects/commit/2c1abf2814e3e4c536c485a1eed1a7789df39e82)

PR #548, opened 2024-01-26
Remove workflow files

volterra79:
Related to https://github.com/g3w-suite/g3w-client-plugin-editing/pull/85


volterra79:
Duplicate #545

PR #576, opened 2024-02-14
Add shortcut link to read the docs within app navbar

Raruto:
## Motivation

Have a quick reference to the current code documentation, eg: https://g3w-suite.readthedocs.io/en/v3.7.x/

## Before

image

## After

image


leolami:
Hi @Raruto, thanks for the suggestion but I don't think it's useful for the following reason

All customers would have the web client with a Help button pointing to the complete suite manual with aspects that the user is not interested in

You have to think about the use of the client by an low leverl user and not by an administrator

It could be useful if the associated link could be customized with links to their specific manuals

But in this case the button should only be visible if the link has been defined

leolami:
Maybe that button is more useful in the ADMIN session

Raruto:
Superseded by: https://github.com/g3w-suite/g3w-client/pull/679

PR #603, opened 2024-03-25
Add support for `showmarker=1` and `iframetype=map` urls parameters

volterra79:
1. **showmarker**  show a marker on center of map. Possible values 1 (show) - 0 (not visible). 

_Ex. ?showmarker=1_

Screenshot from 2024-03-25 15-30-57

3. **iframetype** set type of iframe. Actually only _map_ value is handle. map mean no mapcontrols. Show only map

_Ex: ?iframetype=map_

Screenshot from 2024-03-25 15-36-26


src/app/gui/map/mapservice.js
```diff
@@ -176,18 +176,24 @@ class OlMapViewer {
...
+      if (animate) {
+        view.animate(
+          { duration: 200, center: coordinates },
+          { duration: 200, resolution }
+        );
+      } else {
+        view.setCenter(coordinates);
+        view.setResolution(resolution);
+      }
+    })
```
Raruto, line 199: @volterra79 the else condition will never be reached (ref: `options.animate || true`)

volterra79: @Raruto right. Need to change it. Need to check if boolean otherwise set true

src/app/gui/map/mapservice.js
```diff
@@ -1691,64 +1709,90 @@ proto._resetView = function() {
...
+    };
+
+  (new Promise((resolve) => {
+    setTimeout(async () => {
+      let geom;
+      if (zoom_to_fid) {
+        await this.zoomToFid(zoom_to_fid);
+      } else if (zoom_to_features) {
+        await this.handleZoomToFeaturesUrlParameter({ zoom_to_features });
+      } else if (coords.lat && coords.lon && !Number.isNaN(coords.lat) && !Number.isNaN(coords.lon)) {
```
Raruto: @volterra79 Shouldn't that be an acceptable value?

```
coords.lat = 0
coords.lon = 0
```

src/app/gui/map/mapservice.js
```diff
@@ -1691,64 +1709,90 @@ proto._resetView = function() {
...
+        geom = new ol.geom.Point(ol.proj.transform([coords.lon, coords.lat], 'EPSG:4326', this.getEpsg()));
+      } else if (coords.x && coords.y && !Number.isNaN(coords.x) && !Number.isNaN(coords.y)) {
```
Raruto: @volterra79 Shouldn't that be an acceptable value?

```
coords.x = 0
coords.y = 0
```

volterra79: @Raruto right. Need to check if coords.x && coords.y are not undefined or null. 0, 0 it is acceptable

src/app/gui/map/mapservice.js
volterra79: @Raruto right. Need to check if coords.lat && coords.lon are not undefined or null. 0, 0 it is acceptable

Raruto (approved):

Raruto (approved):

PR #643, opened 2024-06-27
Handle reset of strictly nested dependencies (search inputs)

### LINKED ISSUE #642: Search with dependencies inputs
### Checklist

### Subject of the issue

In case of a search having than 2 strictly dependencies select inputs, changing first input value  reset only first directly input values

### Steps to reproduce

1- Open a search with 3 input strictly dependencies (**Foglio -> Sezione , Particella -> Foglio**) :
 
Screenshot from 2024-06-26 12-33-54

2 - Set a value of Sezione

Screenshot from 2024-06-26 12-34-30

3 - Set a value of Foglio
Screenshot from 2024-06-26 12-35-07

4 - Set a value of Particella 

Screenshot from 2024-06-26 12-35-32

5 - Change a value Sezione

Screenshot from 2024-06-26 12-36-02

Value and select optison value of foglio are changed. Particella is not set to ALL.

### Environment

- g3w-admin: _v3.7.9_
- g3w-client: _v3.9.9_

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes https://github.com/g3w-suite/g3w-client/issues/642

### DIFF
--- a/src/components/SearchPanel.vue
+++ b/src/components/SearchPanel.vue
@@ -224,19 +224,13 @@
        * Sync `this.state.forminputs` with `input.value`
        */
       async changeInput(input) {
-        //get field name
-        const field  = input.attribute;
-        // get parent eventually -> current input dependance
-        const parent = this.state.forminputs.find(i => input.dependance === i.attribute);
-        // inputs
-        // that depend on the current one
-        const deps   = this.state.forminputs.filter(i => field === i.dependance);
+        const field  = input.attribute;                                           // current field name
+        const deps   = this.state.forminputs.filter(i => field === i.dependance); // inputs that depend on the current one
         const state  = this.state;
         let value    = input.value;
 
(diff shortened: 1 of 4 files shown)


src/components/SearchPanel.vue
```diff
@@ -259,31 +253,28 @@
...
             ]));
 
             // value is empty → disable dependants inputs
             d.disabled = is_empty(value) ? d.dependance_strict : false;
 
-            //Check in deep dependencies
-            if (this.state.forminputs
-              .filter(i => i.dependance === d.attribute && (has_select(i) || has_autocomplete(i)))
-              .length > 0
-            ) {
```
Raruto: @volterra79 double check you too, but they should be equivalent.

Raruto (approved):
tested with:

image

src/components/SearchPanel.vue
volterra79: @Raruto i think the problem of deep dependecies is due only if the select input is manage by select 2 directive. ChangeInput is fired if the user change select input type , but not if it changed dynamically

PR #647, opened 2024-07-11
Group all queryby [bbox, drawpolygon, polygon] into a single map control

volterra79:
# Depends on: https://github.com/g3w-suite/g3w-client/pull/646
# Supersedes: https://github.com/g3w-suite/g3w-client/pull/599
# Supersedes: https://github.com/g3w-suite/g3w-client/pull/451

# Before

image

# After

Area controls are grouped under a single button:

## Query by BBOX

image

## Query by Draw Polygon

image

## Query by Polygon

image


Raruto (approved):

src/services/iframe.js
```diff
@@ -274,7 +280,7 @@ class BaseIframeService extends G3WObject {
       PluginsRegistry
         .onafter('registerPlugin', async plugin => {
           await plugin.isReady();
-          if (plugin.getName() === this.pluginName) {
```
Raruto: @volterra79

Yoda conditions should be always written in order to avoid variable assignments. That means, if somewhere you accidentally wrote a single `=` (instead of `===`), your code should crash:

```js
plugin.getName() = this.pluginName // OK → cannot assign to function call
```

```js
this.pluginName = plugin.getName() // KO → valid variable assignment
```

volterra79: @Raruto ok

PR #711, opened 2025-02-05
`font-weight: normal` as default

Raruto:
Comparing with the g3w-admin UI I realized that in here there is an excessive use of `font-weight: bold;`

## Before

image

image

## After

image

image


volterra79:
@Raruto can we close it?

PR #714, opened 2025-02-05
:bug: Get legend from wms layer through internal qgisserver if set internal

volterra79:
In case of wms layer set from G3w-admin as internal :

Screenshot from 2025-02-05 14-42-14

need to get legend from application qgiserver and not get image from external service


moovida (approved):
I tested this on the current dev and it works greatly. It reflects the updated branding section in the documentation part.

PR #742, opened 2025-02-24
🐛 Repeated values found on multi search layers with the same value of a select input

volterra79:
The issue id due when a we create a search with a select input and search from another layer:

Project https://dev.g3wsuite.it/en/map/test-walter/qdjango/296/

Screenshot_20250225_160758

**Before:**

Application search get values, key/value, from each layers. It happen that if layer 1, layer 2 and layer 3 have some item equal, for example [keyA, valueA], these values are show repeated on select input list options:

Screenshot_20250225_161126

**After:**

No name is repeated

Screenshot_20250225_161637

### DIFF
--- a/src/components/CatalogTristateTree.vue
+++ b/src/components/CatalogTristateTree.vue
@@ -155,7 +155,7 @@
 
         <!-- TOGGLE FILTER  -->
         <span
-          v-if                         = "!layerstree.external && (layerstree.selection.active || layerstree.filter.active) && !layerstree.filter.pagination"
+          v-if                         = "!layerstree.external && layerstree.selection.active && layerstree.filter.active && !layerstree.filter.pagination"
           class                        = "action-button skin-tooltip-left selection-filter-icon"
           data-placement               = "left"
           data-toggle                  = "tooltip"
--- a/src/components/CatalogTristateTree.vue
+++ b/src/components/CatalogTristateTree.vue
@@ -155,7 +155,7 @@
 
         <!-- TOGGLE FILTER  -->
         <span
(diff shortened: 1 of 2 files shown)


Raruto (changes requested):
No, this needs to be resolved at server side (response).

src/utils/getDataForSearchInput.js
```diff
@@ -21,7 +21,11 @@ export async function getDataForSearchInput({ state, field, suggest }) {
       })))
     )
       .filter(d => 'fulfilled' === d.status)
-      .reduce((acc, d) => acc.concat(d.value.data || []), []) // uniques by fformatter
+      .reduce((acc, d, i) => 0 === i
+        ? acc.concat(d.value.data || [])                                                       // for first layer get all uninques values 
+        : [...new Set([...(d.value.data || []), ...acc].map(JSON.stringify))].map(JSON.parse), // ensure uniques values (search performed on multiple serach_layers)
+        [] 
+      )
```
Raruto, line 28: Same as: https://github.com/g3w-suite/g3w-client/pull/742#pullrequestreview-2639894111

This must be handled at server side (by @wlorenzetti).

src/components/CatalogTristateTree.vue
```diff
@@ -155,7 +155,7 @@
 
         <!-- TOGGLE FILTER  -->
         <span
-          v-if                         = "!layerstree.external && (layerstree.selection.active || layerstree.filter.active) && !layerstree.filter.pagination"
+          v-if                         = "!layerstree.external && layerstree.selection.active && layerstree.filter.active && !layerstree.filter.pagination"
```
Raruto: @volterra79 

image

Raruto (dismissed):

leolami (approved):

PR #743, opened 2025-02-25
🐛 Download layer features from search with pagination

volterra79:
If we set a search with pagination

Screenshot_20250225_123221

and we try to download all feature from search belong to layer

Screenshot_20250225_123347

we get only visible features

Screenshot_20250225_123543

and not all features from search (10 to 17)

**After**

Screenshot_20250225_123658

we get all 17 features

### DIFF
--- a/src/components/QueryResults.vue
+++ b/src/components/QueryResults.vue
@@ -760,7 +760,7 @@
         this.$options.service.printAtlas(layer);
       },
       showLayerDownloadFormats(layer) {
-        this.$options.service.showLayerDownloadFormats(layer)
+        this.$options.service.showLayerDownloadFormats(layer);
       },
       saveLayerResult(layer, type = "csv") {
         this.$options.service.downloadFeatures(type, layer, layer.features);
--- a/src/services/queryresults.js
+++ b/src/services/queryresults.js
@@ -1624,9 +1624,11 @@ export default new (class QueryResultsService extends G3WObject {
     }
 
     const { query = {} } = this.state;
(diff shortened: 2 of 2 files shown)


Raruto (changes requested):
@volterra79

1. this is not `bug`, **that term should be used sparingly!** (ie. this a `ux` change or a `feature` request)
2. let fronted user decide whether should be preferred behavior (eg. add checkbox/button for: **download all** / **only visible**)

volterra79:
> @volterra79
> 
> 1. this is not `bug`, **that term should be used sparingly!** (ie. this a `ux` change or a `feature` request)
> 2. let fronted user decide whether should be preferred behavior (eg. add checkbox/button for: **download all** / **only visible**)

@Raruto i think that:
1 -  is a bug. Before introduce pagination, user, when clicked on download button , expects all feature of search belong to layer.
2 - @leolami what do you think about?

leolami:
@Raruto 

For now let's fix it by allowing the download of all layers without the visible/all option

Then, maybe, we'll see about proposing it as a feature in 3.10

Raruto:
>  1 - is a bug. Before introduce pagination, user, when clicked on download button , expects all feature of search belong to layer.

@volterra79 Ok, I get the clue.

But please, always add a "short summary" of why you think it's a bug on new issues/PRs you create.

Without that info, the "before/after" section is hard to understand (ie. both the modification and the reason must be understood).

Raruto (approved):

PR #756, opened 2025-03-18
Allow user whether to download layer with relations

volterra79:
Referred to https://github.com/g3w-suite/g3w-admin/issues/1069

Image

It is changed also the case of single download format:

**Before**

_Result content:_

Screenshot_20250318_173311

Tooltip show the download format. No select is showed

_TOC Context Menu:_

Screenshot_20250319_115413

**After**

_Result content:_

Screenshot_20250318_173554

In case of single download, select is show with single option.
It is needed to show also download with relations checkbox 

_TOC Context Menu:_

Screenshot_20250319_114845


src/components/CatalogContextMenu.vue
```diff
@@ -610,7 +620,8 @@
        */
       canDownload(format, layerId) {
         const layer = getCatalogLayerById(layerId);
-        return layer && layer['is' + format + 'Downloadable']();
+        //exclude pdf forma. It is used only for single feature download
```
alexbruy:
```suggestion
        //exclude pdf format. It is used only for single feature download
```

src/components/CatalogContextMenu.vue
```diff
@@ -664,7 +676,10 @@
         ApplicationState.download = true;
         try {
           await getCatalogLayerById(layerId)['get' + format]({
-            data:  map_extent ? { map_extent: GUI.getService('map').getMapExtent().toString() } : undefined
+            data: {
+                    ...(map_extent ? { map_extent: GUI.getService('map').getMapExtent().toString() } : {}),
+                    down_with_relations: Number(this.down_with_relations) // convert boolen to 0/1
```
alexbruy:
```suggestion
                    down_with_relations: Number(this.down_with_relations) // convert boolean to 0/1
```

src/components/QueryResultsActionDownloadFormats.vue
```diff
@@ -74,12 +86,13 @@ export default {
           download,
           this.featureIndex,
           'pdf' === download.format
-           ? document.querySelector(`[feature-html-content="${this.layer.id}_${this.featureIndex}"]`).innerHTML
-           : null
+            ? document.querySelector(`[feature-html-content="${this.layer.id}_${this.featureIndex}"]`).innerHTML
+            : null,
+          Number(this.down_with_relations),  //@since 3.11.7 Convert booelan to 0/1
```
alexbruy:
```suggestion
          Number(this.down_with_relations),  //@since 3.11.7 Convert boolean to 0/1
```

alexbruy (approved):
I'm quite new to the codebase and UI/UX best-practices in the the project, so I may miss something.

In general code looks good to me. There is a couple of minor typos, but the rest is ok.

src/locales/en.js
```diff
@@ -385,6 +385,7 @@ export default {
         back_to_relations: 'Back to relations',
         list_of_relations_feature: 'List of relations of feature',
         error_missing_father_field: "Field is missing",
+        download_with_relations: "Download with relations",
```
alexbruy, line 388: Probably the same string should be added to the rest of i18n files so translators can spot new string and submit translation?

volterra79:
> I'm quite new to the codebase and UI/UX best-practices in the the project, so I may miss something.
> 
> In general code looks good to me. There is a couple of minor typos, but the rest is ok.

@alexbruy thanks

leolami (approved):

PR #758, opened 2025-03-20
Prevent copying static assets from `g3w-admin-*` plugins  (`npm run build`)

volterra79:
Exclude to copy g3w-admin client plugins assests to g3w-admin static client folder when run **npm run build** script  

https://github.com/g3w-suite/g3w-client/blob/b89bfb36abb3187c40ce70f73a654c18321e11bd/gulpfile.js#L48

Below  show qplotly g3w-admin plugin images

Screenshot_20250320_111526


Raruto (approved):

volterra79:
@Raruto need to do a fix. It's seems that images and fonts from g3w-admin-X still copy to admin

PR #762, opened 2025-03-26
Ensure sending `ordering` parameter to server only after user click has clicked on a column (table pagination)

### LINKED ISSUE #761: Sorting records in 1:N related data view
### Checklist

- [x] I've searched through the [current issues](../issues) to make sure this feature hasn't been requested already.

### Motivation

It is necessary that the records (of the child layer) displayed within a 1:N relationship, are ordered according to the order they have in the child table/view

If the user changes the display order, the original order will only be re-presented at a subsequent access to the relations

### Suggested solution

Manage the display order based on the actual sort order of the relation's child table/view

### Alternatives considered

_No response_

volterra79:
Closes: #761

## Before:

In case of pagination, data are get form server with pagination, first pagination data is request without **ordering** parameter. Click on page 2, for example, ordering parameter is passed automatically, first column, without click on column to explicit ordering request.

## After:

**ordering** parameter is in every request to server only when i clicked, for first time, an explicit user ordering 

Screenshot_20250326_111305


src/components/Relation.vue
```diff
@@ -512,6 +520,7 @@
             deferLoading:   data_from_server && this.table.count,
             ajax: data_from_server ? async (opts) => {
               try {
+                this.change_ordering = this.change_ordering || [{ column: 1, dir: 'asc' }].reduce((a, o, i) => a && (o.column !== opts.order[i].column || o.dir !== opts.order[i].dir), true);
```
Raruto: @volterra79 please add a comment, It's a bit unclear what is going here.

Raruto: @volterra79 I'm referring to this part 🧙: `[{ column: 1, dir: 'asc' }].reduce( ...`,

perhaps you meant?

```js
this.change_ordering = this.change_ordering || opts.order.every((o, i) => o.column !== 1 || o.dir !== 'asc');
```

Raruto: or even simpler:

```js
this.change_ordering = this.change_ordering || (1 !== opts.order[0].column || 'asc' !== opts.order[0].dir);
```

volterra79: @Raruto approved. Please change it in this way

Raruto:
@volterra79 Is this expected?

1. open table
2. go to page: `#2`
3. sort by any column (eg. `Type of Maintenance`)
4. table goes to page: `#1`
5. go to page: `#2`
6. sort again by that column (`Type of Maintenance`)
7. table goes to page: `#1`

Raruto:
@leolami Are you sure you want to backport this? (It's not a bug, nor a minor patch)

Keeping adding tons of code into a "stable" branch is a good way to generate bugs 🐞

src/components/Relation.vue
```diff
@@ -339,235 +300,241 @@
...
+              opts.sort = ('desc' === opts.sort ? 'asc' : 'desc');
+            }
+
+            // in case of set ordering (field sort) or not set ordering (start time)
+            if (opts.ordering || opts.order.length) {
+              opts.ordering = `${'desc' === opts.sort ? '-' : ''}${this.table.fields[opts.order[0].column - Number(!!this.showTools)].name}`;
+            }
+
+            /** @type { number } current column index */
+            opts.sort_column = opts.ordering && (this.table.fields.findIndex(({ name }) => ('desc' === opts.sort ? opts.ordering.slice(1) : opts.ordering) === name) + Number(!!this.showTools) ); //need to add 1 if threa are some
```
Raruto, line 363: @volterra79 pleas double check, I think it can be simplified further (ref: `prevOpts`)

volterra79: @Raruto i'll check

volterra79:
> @volterra79 Is this expected?
> 
> 1. open table
> 2. go to page: `#2`
> 3. sort by any column (eg. `Type of Maintenance`)
> 4. table goes to page: `#1`
> 5. go to page: `#2`
> 6. sort again by that column (`Type of Maintenance`)
> 7. table goes to page: `#1`

@Raruto is how datatable works https://datatables.net/examples/basic_init/table_sorting.html

Raruto (approved):

PR #771, opened 2025-04-01
✨ Allow users to search/filter relation table columns

### LINKED ISSUE #770: Filter fields on related table
### Checklist

- [x] I've searched through the [current issues](../issues) to make sure this feature hasn't been requested already.

### Motivation

It would be necessary to allow filtering the related table (1:N) based on the field values, as can be done in general on the attribute table

Image

### Suggested solution

Replicate the functionality as in the standard attribute table

Image

### Alternatives considered

_No response_

volterra79:
Closes: #770 

## Before

Screenshot_20250401_123956

## After

Screenshot_20250401_124023


src/components/Relation.vue
```diff
@@ -315,9 +329,12 @@
 
           // Destroy previous table
           if (this.$table) {
-            this.$table.destroy(true);
-            this.$table     = null;
-            this.table.rows = [];
+            /** @TODO remove "data_from_server" and use table.ajax.reload(); instead */
+            this.$table.destroy(!opts.data_from_server); // In case of data_from_server set .destroy(false), https://datatables.net/reference/api/destroy()
```
Raruto: @volterra79 we will make use [ajax.reload](https://datatables.net/reference/api/ajax.reload) otherwise the input field (search filter) loses keyboard focus if you are too much slow in typing.

Raruto (approved):

src/components/Relation.vue
volterra79: @Raruto i'll check

volterra79: @Raruto i tried. I receive

Screenshot_20250401_145617

and we have an infinite loop

Raruto: > we have an infinite loop

Yes I know, the `DataTable` part should be completely rethought, ie. much more similar to:

https://github.com/g3w-suite/g3w-client/blob/fe580a74abe8ba75f266deff5a68398c18ca16a6/src/components/Table.vue#L701-L716

Furthermore, in future we should also consider replacing the `Relation.vue` component with `Table.vue` (more or less they do the same thing).

src/components/Relation.vue
```diff
@@ -90,72 +87,80 @@
...
-              v-if  = "showTools"
-              class = "table-tools"
+          <tbody id = "table_body_attributes" hidden></tbody>
+          <tbody ref = "table_body">
+            
+            <tr
+              v-for  = "(row, index) in table.rows"
+              :key   = "table.rows_fid[index]"
+              :class = "{
+                'selected': table.rowFormStructure === row,
```
Raruto: @volterra79 is this undefined?

PR #757, opened 2025-03-19
Add `g3w` data provider and refactor `queryby` (area) map control

volterra79:
Referred to https://github.com/g3w-suite/g3w-admin/pull/1071

In case of geometry request by query area map control 

Screenshot_20250319_101959

Screenshot_20250319_101841

we use vector api instead a qgis server

**SERVICE: WFS
VERSION: 1.1.0
REQUEST: GetFeature**

Removed highlight border

Screenshot_20250319_164610

**Using Api, all querable and geometry layers are querable**

Add a new **g3w** provider for Query  area/ geometry request. Leave wfs provider for other use.

PR #774, opened 2025-04-08
Filter selection on query area control + saving of forms with fields with a default expression (without dependencies)

volterra79:
# 🐛 Filter selection on query area

Improves: https://github.com/g3w-suite/g3w-client/pull/757, When do a query selection with Query Area Map control  Screenshot_20250408_154844, selection doesn't work beacuse G3W_FID field in feature is missing

# 💄 Show html in feature body header

In case of feature value contains html code on response, in feature body result header it is show interpreted html

**Before** 

description has paint text

Screenshot_20250408_150500

**After**

Screenshot_20250408_151433

# 🐛  saveDefaultExpressionFieldsNotDependencies

When save current form changes, and fields have default expression without dependencies and expression value is different to current field value, it was waiting a _changeInput_ event. It as wrong because field input with new value is set directly in _getDefaultExpression_

Test: https://dev.g3wsuite.it/en/map/g3w-suite-demo/qdjango/97/ building

### DIFF
--- a/src/components/QueryResultsHeaderFeatureBody.vue
+++ b/src/components/QueryResultsHeaderFeatureBody.vue
@@ -24,7 +24,7 @@
         v-else-if = "isImage(getLayerField({layer, feature, fieldName: attribute.name}))"
         class     = "skin-color"
         :class    = "g3wtemplate.getFontClass('image')"></span>
-      <span v-else >{{feature.attributes[attribute.name]}}</span>
+      <span v-else v-html = "feature.attributes[attribute.name]"></span>
 
     </td>
     <td v-if="!hasLayerOneFeature(layer)">
--- a/src/components/QueryResults.vue
+++ b/src/components/QueryResults.vue
@@ -285,7 +285,7 @@
                 <template v-else-if = "hasFormStructure(layer)">
                   <table class = "table" :class = "{'mobile': isMobile()}">
                     <tbody>
-                      <template v-if = "showFeature(layer, feature)" v-for = "(feature, index) in layer.features">
+                      <template v-for = "(feature, index) in layer.features.filter(f => showFeature(layer, f))">
                         <header-feature-actions-body
                           :colspan                 = "getColSpan(layer)"
                           :actions                 = "state.layersactions[layer.id]"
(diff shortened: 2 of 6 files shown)


Raruto (changes requested):
poor information, you can do better..

src/components/QueryResultsHeaderFeatureBody.vue
```diff
@@ -24,7 +24,7 @@
         v-else-if = "isImage(getLayerField({layer, feature, fieldName: attribute.name}))"
         class     = "skin-color"
         :class    = "g3wtemplate.getFontClass('image')"></span>
-      <span v-else >{{feature.attributes[attribute.name]}}</span>
+      <span v-else v-html = "feature.attributes[attribute.name]"></span>
```
Raruto, line 27: 1. **not a BUG**, and should not be backported (other users can complain)
2. bad UX (seriously, show any type of user submitted html/code?)
3. **links are already handled** in code (a few lines above: [#L12-L16](https://github.com/g3w-suite/g3w-client/blob/994d8be2849a1b6e31abca27d52694d57da3fae1/src/components/QueryResultsHeaderFeatureBody.vue#L12-L16)):
 
image

Raruto (changes requested):

src/components/QueryResultsHeaderFeatureBody.vue
volterra79: @Raruto user can choose to handle a field in html (editing)
Screenshot_20250409_094000
or just use a text input. In this case, if user write a url, it is interpreted has link. 
At the end is the user that define how form inputs are.
@leolami What about this?

leolami:
Hi @raruto, I know it's not a bug but still a different management of the contents between the preview and the extended display of the results.
This change meets some specific requests
We will take care of describing this behavior in the manual and it will be the administrator's responsibility to correctly manage the contents taking this behavior into account

leolami:
Don't just think about the link issue (that's just an example to check the correct HTML interpretation).

In general, it's the possibility to show an html text in the preview

Raruto (approved):
> We will take care of describing this behavior in the manual

💀 death to the manuals..

PR #785, opened 2025-04-14
Missing "active filter" warning within search panel (layer with relations)

### LINKED ISSUE #784: In serach with relations, no filter message is show when run a search with autofilter checked
### Checklist

- [x] I've searched through the existing [issues](https://github.com/g3w-suite/g3w-client/issues) and this bug has never been reported before

### Subject of the issue

In case of a search with relations

Image

After run search 
Image

No message filter set is show

Image

### Steps to reproduce

1. Create a search with relation layer

### Environment

- g3w-admin: v3.9.3
- g3w-client: v3.11.7

### Link to your project

_No response_

### Additional info

_No response_

volterra79:
Closes: #784 

## Before

Screenshot_20250414_145642

## After

Screenshot_20250414_145559


src/components/SearchPanel.vue
```diff
@@ -225,7 +225,7 @@
       },
 
       filterlayers() {
-        return (ApplicationState.tokens.filtertoken && this.state.search_layers.filter(l => l.getFilterToken()) || []);
+        return (ApplicationState.tokens.filtertoken && [].concat(this.state.search_1n_layer || [], this.state.search_layers).filter(l => l.getFilterToken()) || []);
```
Raruto: There are a lot of these statements:

```js
[].concat(state.search_1n_layer || [], state.search_layers)
```

Please try to rewrite it to in order to use only `state.search_layers` and `state.search_1n_relationid`. 

ie. `state.search_1n_layer` is a superfluous property and it just adds complexity.

Raruto (changes requested):

src/components/SearchPanel.vue
volterra79: @Raruto search_1n_layer is not superfluo. Is a layer whe result of search are show. I f you check search configuration, serach_layers are refferred to "parent layers"

src/components/g3w-search.js
```diff
@@ -99,7 +99,7 @@ export function SearchPanel(opts = {}, show = false) {
             : []
           ),          
         ...(input.dependance_strict || 'selectfield' !== input.type || ('selectfield' === input.type && state.child) //@in the case of parent, values are stored
-              ? filtered ? input.values : input.values.filter(v => SEARCH_ALLVALUE !== v.value )// remove `SEARCH_ALLVALUE` from values
+              ? input.values.filter(v => filtered || SEARCH_ALLVALUE !== v.value )// avoid showing a duplicated `SEARCH_ALLVALUE` when removing filter to dependant input
```
volterra79: @Raruto why don't you like ternary conditions? :)

Raruto: @volterra79 @wlorenzetti

> why don't you like ternary conditions? :)

It's not about ternaries, **nested conditions must be avoided** because they make code harder to read and debug (this also applies to long chains of `if-elseif-elseif-...-elseif-elseif-else` conditions).

As a general rule you should always **minimize nesting and prioritize code clarity** (also over any hypothetical performance gains, e.g. `if-elseif` vs `if+if`, the second form is much easier to debug and refactor).

https://www.sonarsource.com/blog/stop-nesting-ternaries-javascript/

src/components/g3w-search.js
```diff
@@ -38,7 +39,7 @@ export function SearchPanel(opts = {}, show = false) {
     search_endpoint:      'api',
     search_1n_relationid: opts.options.search_1n_relationid, //relations
     /** Layers that will be searchable for that search form. The First one is a layer owner of the search set on admin. */
-    search_layers:        [(opts.options || {}).querylayerid || (opts.options || {}).layerid || null, ...((opts.options || {}).otherquerylayerids || [])].map(id => getCatalogLayerById(id)),
+    search_layers:        [(opts.options || {}).querylayerid || (opts.options || {}).layerid || null, ...((opts.options || {}).otherquerylayerids || [])].map(id => getCatalogLayerById(id)).filter(Boolean),
```
Raruto: filtering out invalid layers seems conceptually wrong (layer ids are passed directly from server).

Please show me a case.

Raruto: @volterra79 can you reply on this? (instead of closing it..)

volterra79: @Raruto it could be happen when a search with options:

Screenshot_20250415_155112

referred to a layer that it was removed from project , or hide from TOC

Raruto: @volterra79 i feel that it might be wrong to use `getCatalogLayerById` to get "searchable" layers 

As I said `.filter(Boolean)` hides a potential bug/mismatch from what is saved onto server.

volterra79: @Raruto i use it because it used in other part of code

https://github.com/g3w-suite/g3w-client/blob/56625558444e18844a1330ba2c78859559b8c08f/src/services/queryresults.js#L789

Raruto: > https://github.com/g3w-suite/g3w-client/pull/785#discussion_r2044599051

Please send me a project (or some custom `config.js`) so I can test it with more confidence.

volterra79: @Raruto you can test it setting a search with searching other layers set with a layer

Screenshot_20250415_173707

After that delete the layer from project, using QGIS, and update project on admin. Try to use the search and you'll see an array contain [<LAYER>, undefined]

Raruto: @wlorenzetti this is a server bug, when you re-upload the .qgis project, there is no validation if a layer present in the saved widget gets removed:

image

src/components/g3w-search.js
```diff
@@ -190,22 +212,20 @@ async function doSearch({
...
-        await GUI.closeContent();
-        const opts = (data.data[0] || {}).data;
-        opts.child = true; //set child true
-        //and open a new Search panel
-        new SearchPanel(opts, true)
-      } else {
-        //otherwise, mean the return of search has no values, so we can show an empty results
-        GUI.outputDataPlace(Promise.resolve({ data: [] }));
-        data = [];
-      }
```
Raruto, line 208: @volterra79 please show me how to test this

volterra79: @Raruto you need to install _cadastre_ plugin. It use this type of search. The first search response, return a new search config that is used to build a new search form

leolami (approved):

Raruto (approved):

PR #787, opened 2025-04-17
Layer "filter/selection" within paginated html tables (attribute table + search responses)

volterra79:
## Motivation

In case of paginated searches, not all features are get from response, so:

1. the filter icon should not be visible inside the table (only the selection icon).
2. if there is an active filter results must be automatically selected within the table

## Test case: “filter icon shouldn't be visible within the attribute table”

Perform a paginated search  (with "auto-filter results" option checked)  and then click onto the open attribute table icon (query response):

Screenshot_20250417_105240

### Before

Filter icon is shown. Single feature selection (on each row item) is also possible.

Screenshot_20250417_105343

### After

Filter icon is not shown. Single feature selection (on each row item) is NOT possible.

image

**NB** there is also an active filter (only 10 feature of 13 are retrieved) and results are automatically selected within the table.

---

## Test Case: “keep all features selected when changing page”

Perform a paginated search (with "auto-filter results" option checked) and then:

1. change page size
2. select all features

Screenshot_20250417_112509

### Before

After click on filter icon, only last 3 feature are shown as selected.

Screenshot_20250417_112830

### After

All 13 features are selected. Single feature selection (on each row item) is also possible.

Screenshot_20250417_113015


volterra79:
@leolami @Raruto i think that we need to avoid to **select all features** in Attribute table because if we enable the user to select all feature of a layer, in case of many features (more than 1000 for example),  application will crash. Selection need to be limited to all rows visible at moment in table.

Raruto:
> Selection need to be limited to all rows visible at moment in table.

@volterra79 yeah, I agree (same goes for "pagination" that should also be the default option), but you can't do that in a backport .

An let me be pedantic: this kind of stuffs must not count as bug (has there been a regression? or a fatal error?), but simply an unimplemented feature/improvement.

volterra79:
> > Selection need to be limited to all rows visible at moment in table.
> 
> @volterra79 yeah, I agree (same goes for "pagination" that should also be the default option), but you can't do that in a backport .
> 
> An let me be pedantic: this kind of stuffs must not count as bug (has there been a regression? or a fatal error?), but simply an unimplemented feature/improvement.

@Raruto i agree. Changed PR

Raruto (approved):
I give up 🤷

PR #805, opened 2025-05-23
✨ Share map permalink

volterra79:
Referred to https://github.com/g3w-suite/g3w-admin/pull/1102


Raruto:
Superseded by: https://github.com/g3w-suite/g3w-client/pull/805

PR #810, opened 2025-06-05
Unable to download feature after opening attribute table

volterra79:
## Motivation

If we show a feature click on Form view from attribute table

Screenshot_20250605_124442

and try to download feature

Screenshot_20250605_124545

## Before

Get a javascript error: "Cannot read properties of undefined (reading 'join')"

Screenshot_20250605_124620

## After

Show dialog to save file

Screenshot_20250605_124708

### DIFF
--- a/src/services/queryresults.js
+++ b/src/services/queryresults.js
@@ -1563,16 +1563,25 @@ export default new (class QueryResultsService extends G3WObject {
 
     const { query = {} } = this.state;
 
-    const data = 'search' === query.type && query.search
-      ? { field: query.search.join() }                           // search results + pagination (see: https://github.com/g3w-suite/g3w-client/pull/743)
-      : { fids: query.fids || features.filter(f => !(layer.filter || {}).active || f.selection.selected).map(f => f.attributes[G3W_FID]).join(',') }; // other query types ('point', 'polygon', 'bbox' ..)
-    
-    data.down_with_relations = down_with_relations; 
-  
-    //In the case of pdf type need to add html element
-    if ('pdf' === type) {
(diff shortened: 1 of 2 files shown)


src/services/queryresults.js
```diff
@@ -1563,16 +1563,25 @@ export default new (class QueryResultsService extends G3WObject {
...
+
+      // other query types ('point', 'polygon', 'bbox' ..)
+      fids: 'search' !== query.type || !query.search
+        ? query.fids || features.filter(f => !layer.filter?.active || f.selection.selected).map(f => f.attributes[G3W_FID]).join(',')
+        : undefined,
+
+      // html element (pdf)
+      html: 'pdf' === type ? html : undefined,
+
+    }).filter(([_, v]) => v !== undefined));
```
Raruto, line 1584: @volterra79 Do you confirm that it is superfluous, ie. it work the same way?

```suggestion
    const data = {

      down_with_relations,

      // search results + pagination (see: https://github.com/g3w-suite/g3w-client/pull/743)
      field: 'search' === query.type && query.search
        ? query.search.join()
        : undefined,

      // other query types ('point', 'polygon', 'bbox' ..)
      fids: 'search' !== query.type || !query.search
        ? query.fids || features.filter(f => !layer.filter?.active || f.selection.selected).map(f => f.attributes[G3W_FID]).join(',')
        : undefined,

      // html element (pdf)
      html: 'pdf' === type ? html : undefined,

    };
```

volterra79: @Raruto ok for me

Raruto (approved):

PR #821, opened 2025-07-07
Add Bulgarian translation

suricactus:
Hope this does not conflict much with #818 . I would be happy to hear how I can change the keys of my JSON to match the new pattern, or have this merged before the other one.

See:
2025-07-22 20 55 32 localhost ed9c0f912555

Also added one more string, see #835


Raruto:
Hi @suricactus, I think your PR will be merge later.

> I would be happy to hear how I can change the keys of my JSON to match the new pattern

From https://github.com/g3w-suite/g3w-client/pull/818:

> - start to prefer English sentences as translation keys, eg:
>   - $t('Open Attribute Table') → $t('catalog_items.contextmenu.open_attribute_table')

So you can simply compare your [`bg.js`](https://github.com/g3w-suite/g3w-client/blob/3bcec78bb9fa555845d8d5c70a9b83f51bdc3b9a/src/assets/locales/bg.js) file with the [`it.js`](https://github.com/g3w-suite/g3w-client/blob/db9251b9c4100a0f09d7fa949d046d8d8238d57e/src/assets/locales/it.js) (which is the one with most translated keys).

Raruto:
@suricactus here it is the final [`it.js`](https://github.com/g3w-suite/g3w-client/blob/f48973524ccd96aed0673c6a86059d650e847ae8/src/assets/locales/it.js) (basically no more nested entries) that you can use as a base to translate.

suricactus:
Thanks, will update my translations in the following days.

suricactus:
@Raruto ready for review.  See updated initial comment.

Raruto (approved):

PR #854, opened 2025-09-16
Select all button doesn't mantain its state across multiple queries

volterra79:
## How to reproduce

https://github.com/user-attachments/assets/6b9b98c0-c610-4255-abb1-bb450bae56f4

## Before

After querying again that features, select all button is un-toggled but all features are selected.

Screenshot_20250916_161657

## After

After querying again that features, select all button is toggled and all features are selected.

Screenshot_20250916_161822

### DIFF
--- a/src/components/Table.vue
+++ b/src/components/Table.vue
@@ -329,7 +329,7 @@ export default {
     async inverseSelection() {
       // get all features
       if (!this.getAll) {
-        await this.getFeatures()
+        await this.getFeatures({ formatter: 1 })
       }
       //invert select attribute from current loaded features
       this.state.features.forEach(f => f.selected = !f.selected);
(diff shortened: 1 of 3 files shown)


Raruto:
@volterra79 further investigation needed:

0. https://dev.g3wsuite.it/it/map/demo-310/
1. filter by column field
2. toggle select all checkbox
3. all rows disappears

Instead, it works fine in: https://v310.g3wsuite.it/it/map/demo-310/

---

image

---

image

volterra79:
> @volterra79 further investigation needed:
> 
> 0. https://dev.g3wsuite.it/it/map/demo-310/
> 1. filter by column field
> 2. toggle select all checkbox
> 3. all rows disappears
> 
> Instead, it works fine in: https://v310.g3wsuite.it/it/map/demo-310/
> 
> image
> 
> image

Raruto (approved):

PR #856, opened 2025-09-30
🐛 Compare relation feature filter expression value with key to select option  [editing]

volterra79:
## Test page

https://v310.g3wsuite.it/it/map/epsg-25833/qdjango/323/

## Before

Try to edit a relation _elementi_caratterizzanti_ from _piani_

Screenshot_20250930_103257

**lin_inq_tsubinq_id** contain a value with parenthesis, mean no value on select values is found.
This because relations  _elementi_caratterizzanti feature from piani (parent layer) feature are get using formatter 1 parameter:

Screenshot_20250930_103611

so we need to compate this value with keys and not value of select options

Screenshot_20250930_103652

## After

Screenshot_20250930_104755

### DIFF
--- a/src/map/layers/layer.js
+++ b/src/map/layers/layer.js
@@ -206,7 +206,7 @@ const Providers = {
             response = await XHR.post({
               url:         createRelationsUrl(options.filter.fid),
               contentType: 'application/json',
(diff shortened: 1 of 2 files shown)


Raruto:
@volterra79 side note, it seems in there datatable rows are very tall:

image

leolami (approved):

Issue #861, opened 2025-10-13
Annotation: text color as geometry color


leolami:
The color of the text associated with a geometry (circle, square...) must have same color of the geoemtry

PR #862, opened 2025-10-13
💄 Change annotation text color

volterra79:
Closes: #861 

## Before

Text are black

Screenshot_20251013_084440

## After

Text are in the same color of annotation geometry

Screenshot_20251013_084521


Raruto:
@leolami I don't recommend it.

It introduces a11y issues. Most of the time, text will be unreadable.

Black provides best color contrast.

leolami:
right @Raruto but f the writing is all black you can get confused when there are multiple objects overlapping

One solution could be to use the same color as the geometry for the text but a little darker.

Raruto (approved):

leolami (approved):

PR #887, opened 2026-02-17
✨Reset map theme when layer/ group change

volterra79:
When a map theme is set

Screenshot_20260217_095430

If we apply change on TOC layer/group (check/uncheck)

Screenshot_20260217_095605

or change layer style

Screenshot_20260217_095649

**Before**

Map theme doesn't reset

Screenshot_20260217_095732

**After**

Map theme is reset

Screenshot_20260217_095832


volterra79:
Move it on milestone 4.1.0

PR #908, opened 2026-03-13
Load vue.min.js version library

volterra79:
Use vue production version

**Before** 

Screenshot_20260313_175140


Raruto (commented):
I wouldn't worry about this (it's just a warning); the final code is optimized by esbuild anyway.

For debugging, it's better to keep "type" checks enabled on production sites.

PR #915, opened 2026-04-07
Run default expression on fields on update with no field dependencies

volterra79:
Change value with default expression filed on update, only if some other fields are change
This is necessary because otherwise in case, for example i change relation features of parent features without any change of parent feature fields, after save changes, it call every time server default expression also with no changes


volterra79:
Move logic to editing plugin

PR #924, opened 2026-04-27
Make use of `ResizeObserver` instead of `"resize"` and `"transitionend"` events

volterra79:
This update replaces the window resize event and main sidebar transitionend event with ResizeObserver to enhance layout management.

## BEFORE

https://github.com/user-attachments/assets/e442a158-f2ab-4159-8071-1ffee5ec6fe3

## AFTER

https://github.com/user-attachments/assets/8f9e77e2-f35e-4236-b2fe-8dea5a674408

### DIFF
--- a/src/g3w-app.js
+++ b/src/g3w-app.js
@@ -668,10 +668,10 @@ export default new (class GUI extends Emitter {
     ApplicationState.sidebar.width = document.querySelector('.main-sidebar').offsetWidth;;
 
     // handle main window resize
-    window.addEventListener('resize', () => { requestAnimationFrame(() => { this._layout(); }); });
+    new ResizeObserver(debounce(() => this._layout(), 0)).observe(document.body);
 
     // handle main sidebar resize
-    document.querySelector('.main-sidebar').addEventListener('transitionend', () => { requestAnimationFrame(() => { this._layout(); }); });
+    new ResizeObserver(() => this._layout()).observe(document.querySelector('.main-sidebar'));
 
     this._layout();
 
--- a/src/g3w-app.js
+++ b/src/g3w-app.js
@@ -667,11 +667,11 @@ export default new (class GUI extends Emitter {
 
     ApplicationState.sidebar.width = document.querySelector('.main-sidebar').offsetWidth;;
 
-    // handle main window resize
-    new ResizeObserver(() => this._layout()).observe(document.body);
-
-    // handle main sidebar resize
-    new ResizeObserver(() => this._layout()).observe(document.querySelector('.main-sidebar'));
+    // handle main window resize and main sidebar resize (shared frame to avoid multiple _layout() calls per frame)
+    let _layoutRaf;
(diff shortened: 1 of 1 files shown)


src/g3w-app.js
```diff
@@ -668,10 +668,10 @@ export default new (class GUI extends Emitter {
     ApplicationState.sidebar.width = document.querySelector('.main-sidebar').offsetWidth;;
 
     // handle main window resize
-    window.addEventListener('resize', () => { requestAnimationFrame(() => { this._layout(); }); });
+    new ResizeObserver(debounce(() => this._layout(), 0)).observe(document.body);
```
Raruto: @volterra79 why `delay = 0`?

volterra79: @Raruto otherwise debounce function has default wait time 600 ms and if you try to resize window, you can see delay for example on "sync" map controls

Raruto: @volterra79 so why do you use the debounce function? (0 = no debounce)

Raruto: @copilot

volterra79: @Raruto need to avoid multi call of this._layout() on resize.

Raruto: @copilot should we use `requestAnimationFrame`? instead of `debounce`? (ie. to avoid multiple invokation from `ResizeObserver`?)

volterra79: @Raruto 

using

` new ResizeObserver(() => requestAnimationFrame(() => { console.log('requestAnimationFrame'); this._layout()})).observe(document.body);`
    
  If you try to resize window, you can see multi console log output for just one resize window move

volterra79: @Raruto 

```
let _layoutRaf;
    const scheduleLayout = () => { cancelAnimationFrame(_layoutRaf); _layoutRaf = requestAnimationFrame(() => {console.log('layout'); this._layout(); }); };
    new ResizeObserver(scheduleLayout).observe(document.body);
```

Screenshot_20260505_111152

just resize twice time window

If you try to close sidebar

Screenshot_20260505_111406

Raruto: @volterra79 even so, it is printed several times:

```js
const _layout = debounce(() => { console.log('layout'); this._layout() }, 0);
new ResizeObserver(_layout).observe(document.body);
new ResizeObserver(_layout).observe(document.querySelector('.main-sidebar'));
```

volterra79: @Raruto right.  Set 100 instead of 0

Raruto:
@volterra79 ref: https://github.com/g3w-suite/g3w-client/pull/924/commits/afb1be903da1d43cd2b8fbfc44ed9262b6fbf3b1

## BEFORE

https://github.com/user-attachments/assets/e442a158-f2ab-4159-8071-1ffee5ec6fe3

## AFTER

https://github.com/user-attachments/assets/8f9e77e2-f35e-4236-b2fe-8dea5a674408

Raruto (approved):

PR #934, opened 2026-05-22
Keep map at 100% width while toggling sidebars above it (floating panels)

volterra79:
### Before

Map resize

Screenshot_20260522_120743

### After

map still 100% width

Screenshot_20260522_120621


src/directives/v-t-tooltip.js
```diff
@@ -80,7 +80,7 @@ const tooltip = Object.assign(document.createElement('template'), {
 document.querySelector('#app').insertAdjacentElement('afterend', tooltip);
 document.addEventListener('mousemove', showTooltip);
 document.addEventListener('mousedown', showTooltip);
-document.addEventListener('focusin', showTooltip);
+document.addEventListener('mouseover', showTooltip);
```
Raruto: @volterra79 what about keyboard navigation? (tab key)

volterra79: Fix with be6d447576f39af1de68b3fd86e20b857b346b62
Now we need to fix javascript issue after toggle Annotation ma control

Screenshot_20260901_170651

volterra79: @Raruto solved with 96e2fd68f697fc69093d04d647a2dc51ffdd9432

Raruto (approved):
It seems pretty good to me.

PR #939, opened 2026-06-03
On mobile, layer tree checkboxes are always unchecked (empty box)

volterra79:
Checkboxes are show as white box when checked

### Before

WhatsApp Image 2026-06-03 at 15 46 28

### After

image


Raruto:
@volterra79 how can I replicate this bug? (browser/device)

volterra79:
> @volterra79 how can I replicate this bug? (browser/device)
On mobile visit:

https://dev.g3wsuite.it/en/map/g3w-suite-demo/qdjango/97/

Raruto (approved):

PR #956, opened 2026-06-18
🐛 Show search title if set instead name

volterra79:
Project https://dev.g3wsuite.it/en/map/g3w-suite-demo/qdjango/97/

Screenshot_20260618_151309

### Before

Screenshot_20260618_151335

Screenshot_20260618_151407

### After

Screenshot_20260618_151427

Screenshot_20260618_151449

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -20,7 +20,7 @@
       
     >
       <i aria-hidden = "true" class = "far fa-circle"></i>
-      <span>{{ search.name }}</span>
+      <span>{{ search.options.title || search.name }}</span>
     </li>
 
     <li v-for = "searchtool in state.tools">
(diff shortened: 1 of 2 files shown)


leonardolami (approved):

Raruto:
@volterra79  Why the client needs to both a search `title` and `name`?

volterra79:
> @volterra79 Why the client needs to both a search `title` and `name`?

Because name, in server configuration is mandatory, instead title is optional. In case title is leave empty, we get name

Screenshot_20260622_084901

Raruto:
@volterra79 this violates SRP and needs to be fixed on the g3w-admin side.

Backend API must handle the fallback logic and return a single, ready-to-use title property to the client, rather than forcing the frontend to evaluate `opts.options?.title || opts.name`.

volterra79:
> @volterra79 this violates SRP and needs to be fixed on the g3w-admin side.
> 
> Backend API must handle the fallback logic and return a single, ready-to-use title property to the client, rather than forcing the frontend to evaluate `opts.options?.title || opts.name`.

I know but for the now, v3.11 g3w-suite release, we need to leave it as it is.

Raruto:
@volterra79 Let me know 👉 https://github.com/g3w-suite/g3w-admin/pull/1422

Raruto:
Superseded by: https://github.com/g3w-suite/g3w-admin/pull/1422

